Issue #32, opened 2016-05-06
Add link to Lets Encrypt on the page


SwartzCr:
One that links directly to letsencrypt.org and one for a donation page

PR #91, opened 2016-05-12
Fix css bugs and tagline link

vbrown608:
-Link tagline to letsencrypt website - #32 
-Increase line height of tagline on mobile
-Fix a bug where the instruction widget dropdown menus don't appear on very short viewports (ex. mobile landscape)


pde:
LGTM

PR #112, opened 2016-05-23
Update package names for Gentoo

propyless:
update package names to reflect those changes done in the gentoo portage tree

Also there is no package called ` {certbot,letsencrypt}-nginx`

```
$ eix certbot
* app-crypt/certbot
     Available versions:  ~0.6.0 **9999 {test PYTHON_TARGETS="python2_7"}
     Installed versions:  0.5.0(21:35:42 19/04/16)(-test PYTHON_TARGETS="python2_7")
     Homepage:            https://github.com/certbot/certbot https://letsencrypt.org/
     Description:         Let's encrypt client to automate deployment of X.509 certificates

* app-crypt/certbot-apache
     Available versions:  **9999 {test PYTHON_TARGETS="python2_7"}
     Homepage:            https://github.com/certbot/certbot https://letsencrypt.org/
     Description:         Apache plugin for certbot (Let's Encrypt Client)

Found 2 matches
```

### DIFF
--- a/_scripts/instruction-widget/install.js
+++ b/_scripts/instruction-widget/install.js
@@ -109,15 +109,12 @@ module.exports = function(context) {
   gentoo_install = function() {
     template = "gentoo";
 
-    context.package = "letsencrypt";
-    context.base_command = "letsencrypt";
-    context.base_package = "app-crypt/letsencrypt";
+    context.package = "certbot";
+    context.base_command = "certbot"
+    context.base_package = "app-crypt/certbot";
     context.cbauto = false;
     if (context.webserver == "apache") {
-      context.package = "letsencrypt-apache";
-    }
-    if (context.webserver == "nginx" && context.advanced) {
-      context.package = "letsencrypt-nginx";
+      context.package = "certbot-apache";
     }
   }
 


SwartzCr:
Looks great, except that I'd also change the base_command to be certbot

_scripts/instruction-widget/install.js
```diff
@@ -109,15 +109,12 @@ module.exports = function(context) {
   gentoo_install = function() {
     template = "gentoo";
 
-    context.package = "letsencrypt";
+    context.package = "certbot";
     context.base_command = "letsencrypt";
```
SwartzCr: To clarify: since the command that you'll be running from the command line will be certbot and not letsencrypt (presumably). So I'd change this line to say certbot as well, so that way the instructions that are generated will say something like
`$ certbot certonly -d www.example.com`
and so on, as opposed to
`$ letsencrypt certonly -d www.example.com`

propyless:
Hello,
The currently available (unstable) ebuild is for version 0.6.0.
If you run the 9999 ebuild (latest HEAD found for certbot in github) then you get 0.7.0.

Since they haven't updated the ebuild for 0.7.0 which provides the certbot command, they just moved and renamed letsencrypt to certbot without updating the release version.

However, if you think that its preferable to update the base command, I can do so :)

SwartzCr:
while 0.6.0 supports the certbot name, it's about how gentoo manages this. With the 0.6.0 package does the commandline command certbot work as an invocation?

propyless:
Ah right, will fix then!
Yes, just tested, it works.

propyless:
Fixed. Thanks for your help!

SwartzCr:
travis is broken, but I'm going to merge this anyway?

propyless:
Thank you ! :)

PR #115, opened 2016-05-24
update package names for fedora

SwartzCr:
https://bodhi.fedoraproject.org/updates/?packages=certbot


SwartzCr:
@nbebout can you confirm that this is correct?

hogarthj:
Yeah this is what we have in our repo now, same for CentOS 7 .

PR #124, opened 2016-06-02
_docs.sh should not run with sudo; use certbot install routine to install virtualenv

mfb:


mfb:
@vbrown608 or @SwartzCr this is pretty trivial and I think ok to merge

SwartzCr:
I'm a little worried about the removal of the pip install virtualenv line - can you explain that?

mfb:
afaik it can be removed because `letsencrypt-auto --os-packages-only` will install virtualenv for you via apt-get (or pip on macOS)

PR #133, opened 2016-06-14
Fix typo

pde:


SwartzCr:
hey peter, your docs directory seems to be out of date. Notably it's missing the move from le_util.rst to util.rst which is causing a test to break. Maybe you need to do a submodule fetch before committing?
and make sure that you have this commit in the tree that you commit: https://github.com/certbot/certbot/commit/b05258243a2b367632c207f3d2a14cc924258225

PR #134, opened 2016-06-15
add note about debian cronjob

### LINKED ISSUE #126: Debian cron job
The Debian package provides a cron job to automatically renew your certificates. Information about this should be included in the Debian instructions.

SwartzCr:
fixes #126

### DIFF
--- a/_scripts/instruction-widget/templates/install/debian.html
+++ b/_scripts/instruction-widget/templates/install/debian.html
@@ -12,9 +12,11 @@
 $ sudo apt-get install {{package}} {{backports_flag}}
 </pre>
 
+{{#above_4}}
 <aside class="note">
     <h4>Note</h4>
     <p>
-    The Debian packaged version of Certbot installs a cron job automatically into /etc/cron.d/certbot. This cronjob runs a renewal twice a day (0000, 1200) with a small offset. Note that this won't renew certificates that aren't within the renewal window, but you may still want to replace this with your own renewal workflow.
+    The Debian packaged version of Certbot installs a cron job automatically into /etc/cron.d/certbot. This cronjob runs certbot renew twice a day which will renew certificates that are within the renewal window.
     </p>
 </aside>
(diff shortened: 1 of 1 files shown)


_scripts/instruction-widget/templates/install/debian.html
```diff
@@ -11,3 +11,10 @@
 <pre>
 $ sudo apt-get install {{package}} {{backports_flag}}
 </pre>
+
+<aside class="note">
+    <h4>Note</h4>
+    <p>
+    The Debian packaged version of Certbot installs a cron job automatically into /etc/cron.d/certbot. This cronjob runs a renewal twice a day (0000, 1200) with a small offset. Note that this won't renew certificates that aren't within the renewal window, but you may still want to replace this with your own renewal workflow.
```
bmw: > This cronjob runs a renewal twice a day (0000, 1200) with a small offset.

I think "runs a renewal" is a strange way to phrase it. Maybe just "This cron job runs `certbot renew` twice a day..."? 

I'm also personally not a fan of the "(0000, 1200)" and think the offset is irrelevant for most people, but if you like it, feel free to leave those parts as is.

> Note that this won't renew certificates that aren't within the renewal window...

How about something like "This will only renew certificates as they approach expiration" or even "This will only renew certificates that are in the renewal window"? Personally, not a fan of the double negative, but again, feel free to leave this as is if you prefer.

bmw:
This note also shows up in the instructions for Xenial which doesn't include a cron job.

pde:
I don't think people should need to replace the Debian renewal script with their own renewal workflow. If `certbot renew --dry-run` works for them, so should the cron job...

SwartzCr:
please suggest edits, and I'll update it, and remove it from the xenial instructions

SwartzCr:
okay it's updated, and doesn't show for ubuntu

bmw:
LGTM

PR #143, opened 2016-06-29
webroot

### LINKED ISSUE #119: Interactive plugin selection not available in Xenial
The instructions for Xenial (for example with Nginx) mention that `letsencrypt certonly` "[...] will allow you interactively select the plugin and options used to obtain your certificate."

That feature isn't available in the version of `letsencrypt` in Xenial - the client will simply use standalone.

The Xenial instructions should probably use `--webroot` by default, while mentioning that `--standalone` is available in case anyone would rather use that for some reason.

I remember reading about plans for a certbot PPA for Xenial, so I guess if that's going to happen soon-ish, it might be easier to just replace the instructions with ones for the PPA, which would make this issue go away as well.

Related topic from the community forum: https://community.letsencrypt.org/t/letsencrypt-renew-does-not-work-ubuntu-nginx/16219

SwartzCr:
fixes #119


_scripts/instruction-widget/templates/getting-started/webroot.html
```diff
@@ -0,0 +1,20 @@
+<p>
+Certbot supports a number of different “plugins” that can be used to obtain and/or install certificates.
+
+Since your server architecture doesn't yet support automatic installation you should use the <a href="https://certbot.eff.org/docs/using.html#webroot">webroot command</a> to obtain your certificate.
```
bmw: Just because we can't automatically install their cert doesn't necessairly mean they should use `webroot`. They may instead want to use `standalone`.

While most of this file looks good, I think we should have wording here closer to the [original certonly.html file](https://github.com/certbot/website/blob/d9c6d84bdbd67f62d775db6084a9c1f22933e194/_scripts/instruction-widget/templates/getting-started/certonly.html) before I messed it up by suggesting they use the UI.

SwartzCr:
added the suggested changes, @bmw can you look it over again?

_scripts/instruction-widget/templates/getting-started/webroot.html
```diff
@@ -0,0 +1,35 @@
+<p>
+Certbot supports a number of different “plugins” that can be used to obtain and/or install certificates.
+</p>
+<p>
+Since your server architecture doesn't yet {{officially}} support automatic
+installation {{imperative}} use use the certonlt command to obtain a cert.
```
bmw: s/certonlt/certonly

_scripts/instruction-widget/templates/getting-started/webroot.html
```diff
@@ -0,0 +1,35 @@
...
+installation {{imperative}} use use the certonly command to obtain a cert.
+</p>
+<p>
+To obtain a cert using the <a href="https://certbot.eff.org/docs/using.html#webroot">webroot</a> plugin, which can work with the webroot directory of any webserver software, run this command.
+</p>
+
+{{#xenial}}
+<aside class="note">
+    <h4>Note:</h4>
+    <p>Ubuntu Xenial has an old version of Certbot packaged for it that lacks a few features. For example the certonly option isn't available, and the <a href="https://certbot.eff.org/docs/using.html#webroot">webroot plugin</a> doesn't have an interactive mode. You can still get a certificate but you'll need to specify all of your variables on the command line.
```
bmw: > For example the certonly option isn't available

`certonly` is available on Xenial.

> You can still get a certificate but you'll need to specify all of your variables on the command line.

Unless there is another feature we're missing in Xenial you'd like to mention, what if we just combined this and the previous sentence to say something like:

"The webroot plugin doesn't have an interactive mode, so you'll need to specify your domains and webroot paths on the command line to use the plugin."

bmw:
One last thing is now that this file talks about both webroot and standalone, maybe we shouldn't call the file and the corresponding javascript functions "webroot".

In fact, maybe we want this to be a special case of `certonly.html`. There are a couple benefits to this:

1. People on Xenial get renewal instructions
2. People on other OS's get the note about `.well-known` needs to be served by your webserver to use standalone.

If we do this, we'll need to make sure that all instructions shown work on Xenial (I think they do, but I didn't check closely). You can use inverted section openers like those used [here](https://github.com/certbot/website/blob/master/_scripts/instruction-widget/templates/install/centos.html#L11) to make this easier.

bmw:
LGTM

PR #144, opened 2016-07-04
Update package name for FreeBSD

patf:
The package name was changed to py27-certbot recently:
https://www.freshports.org/security/py-certbot/

The base command for all FreeBSD packages is now "certbot".

Tested on a clean image of FreeBSD 10.3.


bmw:
LGTM. Thanks @patf.

PR #146, opened 2016-07-11
Fix narrative flow in the case where backports are necessary

pde:


pde:
Trivial change, self-merging.

PR #150, opened 2016-07-14
Don't suggest the apache plugin on *BSD

patf:
According to http://svnweb.freebsd.org/ports/head/security/py-certbot/pkg-message?view=markup, FreeBSD will be getting a separate `security/py-letsencrypt-apache` package in the future, but the apache plugin is not currently included.

OpenBSD packaging plans are unclear.


pde:
LGTM

PR #160, opened 2016-09-21
add xenial warnings

### LINKED ISSUE #159: Include --agree-tos for --dry-run on Xenial
On Ubuntu 16.04, `letsencrypt renew --dry-run` won't work unless you also include `--agree-tos` on the command line. Additionally, user's will see:
```
2016-04-02 13:49:45,115:WARNING:letsencrypt.client:Registering without email!
```
We should tell users to add the `--agree-tos` flag when testing their renewal setup and perhaps also tell them to ignore the e-mail address warning.

SwartzCr:
fixes #159

### DIFF
--- a/_scripts/instruction-widget/templates/getting-started/renewal.html
+++ b/_scripts/instruction-widget/templates/getting-started/renewal.html
@@ -10,7 +10,7 @@ by running this command:
 
 {{#xenial}}
 There's a bug in the version of Certbot on Ubuntu Xenial which may show a warning
-saying 'Registering without email!' evn if you have previously given an email to
+saying 'Registering without email!' even if you have previously given an email to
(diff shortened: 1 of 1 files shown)


_scripts/instruction-widget/templates/getting-started/renewal.html
```diff
@@ -5,9 +5,15 @@
 by running this command:
 
 <pre>
-{{base_command}} renew --dry-run
+{{base_command}} renew --dry-run {{#xenial}}--agree-tos{{/xenial}}
 </pre>
 
+{{#xenial}}
+There's a bug in the version of Certbot on Ubuntu Xenial which may show a warning
+saying 'Registering without email!' evn if you have previously given an email to
```
bmw: s/evn/even

PR #162, opened 2016-10-13
Reorder Dockerfile for speed and fix Travis tests

bmw:
Move things like building `ruby` and installing `jekyll` and `gulp` higher in the `Dockerfile` so they can be cached between website updates.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ addons:
       - texlive-latex-extra
 
 before_install:
-  - rvm install 2.2.1
+  - rvm install 2.2.2
 
 install:
   - npm install gulp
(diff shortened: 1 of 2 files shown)


bmw:
It appears `html-proofer` indirectly requires `ruby >= 2.2.2`. We already install this in the Dockerfile (without this PR), but this makes the change for `.travis.yml` as well.

bmw:
@pde, you should probably review this PR first as every other open PR is built on it (because otherwise, Travis tests wouldn't pass). Running a local version of the website in Docker is now remarkably sane. If you'd like help with this when testing/reviewing these PRs, let me know.

PR #166, opened 2016-10-26
Remove/fix EPEL instructions on CentOS/RHEL 6

bmw:
~~Blocked on certbot/certbot#3651 to pass tests.~~

Built on #162.

I used the id "centosrhel6" for RHEL 6 to prevent breaking links like https://certbot.eff.org/#centosrhel6-other. For these links, it is better to provide instructions for enabling EPEL even if `certbot-auto` can do it for you (CentOS 6 users) than to provide no instructions on how to do it when you have to do it manually (RHEL 6 users).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ addons:
       - texlive-latex-extra
 
 before_install:
-  - rvm install 2.2.1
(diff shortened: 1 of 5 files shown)


pde:
LGTM!

PR #172, opened 2016-11-02
Nonunix

### LINKED ISSUE #147: Non-UNIX/Other OS
We should add another entry to the OS dropdown titled something like "non-UNIX" or "other" to help people on systems like Windows. Unfortunately, Certbot won't help them, but perhaps we can point them to [3rd party software](https://github.com/certbot/certbot/wiki/Links#windows) that will.

bmw:
Fixes #147.

Built on #162 so tests pass.

Please pay special attention to the language I used when recommending ACME clients other than Let's Encrypt.


pde:
LGTM.

PR #173, opened 2016-11-04
Shared hosting instructions

### LINKED ISSUE #137: Help non-sysadmins
As suggested by @schoen, in the instruction generator, we should provide option(s) such as "shared hosting" or "hosting company" or even "I'm not sure" or "I don't know". The reason for this is some people attempt to use Certbot with very limited access to their server. The client provides very little support for this (you have to use the manual plugin and install the certificate by hand every 3 months).

In the text for this option, we should explain the problem to the user and tell them to see if their hosting provider has integrated Let's Encrypt support. If they haven't, perhaps we should suggest they ask their hosting provider to do so. We could also potentially tell the users about the other options available to set up TLS on their site, however, since this isn't likely to be one of our products, we have little/no control over the quality, security, etc of such a product and we should be careful.

bmw:
Fixes #137.

### DIFF
--- a/_scripts/instruction-widget/data/inputs.json
+++ b/_scripts/instruction-widget/data/inputs.json
@@ -23,6 +23,12 @@
     ],
     "operating_systems": [
         {
+            "name": "Web Hosting Service",
+            "id": "sharedhost",
+            "distro": "sharedhost",
+            "version": "0"
+        },
+        {
             "name": "Debian 7 (wheezy)",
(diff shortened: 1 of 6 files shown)


pde:
LGTM; I think this would also make sense as an answer for "I don't know".

pde:
We decided the "I don't know" case was complicated enough to be a separate ticket, if we do it at all.

PR #182, opened 2016-12-12
Remove use-certbot page

vbrown608:
This probably shouldn't be here

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "description": "Gulp build certbot.",
   "scripts": {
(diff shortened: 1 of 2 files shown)


vbrown608:
Looks like this is failing because the link to edit on Github for https://certbot.eff.org/docs/api/log.html 404s. I'm guessing that's because that rtd page doesn't have any content.

I can add it to the list of pages html-proofer should ignore, but it might be better to just stop checking external links for 404s.

bmw:
LGTM! Thanks @vbrown608.

PR #187, opened 2017-01-17
Fix install instructions for OpenBSD

abduelhamit:


_scripts/instruction-widget/install.js
```diff
@@ -161,7 +161,7 @@ module.exports = function(context) {
     }
     if (context.distro == "opbsd"){
       context.portcommand = "letsencrypt/client";
-      context.package = "pkg_add letsencrypt";
+      context.package = "pkg_add certbot";
       context.base_command = "letsencrypt";
```
bmw: Just to verify, is the command also called `certbot` now? If so, this line needs to be changed as well, however, I'm happy to do this myself in a follow up PR if you prefer.

bmw:
This was fixed in #214. Regardless, thanks for the PR!

PR #190, opened 2017-01-20
Don't try to copy use-certbot into Docker

bmw:
This file was removed in #182.


bmw:
@SwartzCr can you take a look at this? I've just been keeping this change locally so I can build releases.

SwartzCr:
yup - lgtm

PR #194, opened 2017-02-16
use preferred piwik URL

mfb:
some adblockers block piwik.php so we are using an alternate path for anon-stats.eff.org


mfb:
anyone have a moment to review & merge this?

mfb:
p.s. tests are actually passing now

PR #195, opened 2017-02-17
Fixed issue #3544

muhammednagy:
I was trying to install certbot on RHEL7 but I had broken dependencies.i googled the error and I found that there already an open issue.
I edited the installation guide 
thanks

### DIFF
--- a/_scripts/instruction-widget/templates/install/centos.html
+++ b/_scripts/instruction-widget/templates/install/centos.html
@@ -6,6 +6,7 @@ Certbot, you must first <a
 href="https://fedoraproject.org/wiki/EPEL#How_can_I_use_these_extra_packages.3F">
 enable the EPEL repository</a>.
 </p>
+<p>Don't forget to enable epel optional channel if you are using RHEL</p> 
 <p>
 After doing this, you can install Certbot by running:
 </p>


muhammednagy:
why nobody replied?`

SwartzCr:
Oh sorry about that! Thanks for the edit @muhammednagy 
I'm curious if you can tell me how this is different than the existing note? Right above it the rhel instructions currently say: 
"Certbot is packaged in EPEL (Extra Packages for Enterprise Linux). To use Certbot, you must first enable the EPEL repository"

muhammednagy:
I had a problem on RHEL when I was trying to install certbot on RHEL.
I enabled epel and tried to install but I had problems 
I searched and I found that there are dependencies  in the optional channel which isn't enabled by default when you enable epel

SwartzCr:
Interesting! Can you link me to some of the resources that you found? I'd like to make sure we explicitly call out the specific dependencies when we document this. Also if you still have your error output that'd be really helpful too.
(and again, thanks for bringing this up - sorry that I missed it for so long :grimacing: )

muhammednagy:
No problem :)
i'm sorry i don't have the error now 
i found the package it was  python-zope-interface
i can edit my add a new pull request and add the package name :)

SwartzCr:
I'm going to tag https://github.com/certbot/certbot/issues/3544 here and see if @hogarthj @chriszrc or @akadata can help with deciding on wording. I'm happy to merge this so long as the wording is right and actually helpful

chriszrc:
It's not a bad start, but as I mentioned here, for people without redhat licenses, it's not even clear how to enable the optional channel. Including some of what I mentioned here might be nice:

https://github.com/certbot/certbot/issues/3544#issuecomment-282064222

muhammednagy:
I was trying on ec2 too it worked this way.
I may add a new section to mention how to enable optional channel on ec2 or add the thread link

SwartzCr:
I think it's fine to add the specific instructions as mentioned, we can update them if they change (where as the thread might get lost to time) @muhammednagy once you do that I'd be happy to review and merge

muhammednagy:
done #208

SwartzCr:
Okay I'm gonna close this in faor of #208

PR #196, opened 2017-03-09
FAQ update: Answer for CSR is no longer misleading.

lancedolan:
Previously, the answer in the FAQ misleads a reader to believe the certbot will accept a private key as input, and that providing a CSR as input is optional. In fact, providing CSR is the only means of using your own private key.


SwartzCr:
Awesome, thanks for the update @lancedolan !

PR #214, opened 2017-04-26
fix issues with openbsd

### LINKED ISSUE #213: Update OpenBSD instructions
Depending on the version, we can tell people to use certbot instead of letsencrypt. There's a courtesy symlink so it's not *wrong* to say letsencrypt, but we don't have to. No apache or nginx packages yet though.

5.8: no package
5.9: letsencrypt 0.3.0
6.0: certbot 0.8.1
6.1: certbot 0.12.0
development: certbot current version (0.13.0)

Releases are every 6 months.

Also, the maintainer recommends that we drop references to `ports`, and only use `pkg_add`:

> > > We generally regard the ports tree as something more for ports developers than typical users, so it might be better to show the pkg_add example first (and possibly even skip the ports one).

> > Would the ports tree have newer versions, if people wanted those? Does
> > it make sense to keep the reference to ports so people can get the
> > latest Certbot, or is it too complex to install from ports?

> In some cases (usually only for security issues) updates or backports might
> be committed to the -stable ports tree for the most recent release (and very
> rarely for the one previous).
> 
> The development ports tree which has the usual version updates is only
> meant for use with recent development snapshots of the OS, sometimes an
> update from here will be OK on the last release but users knowledgeable
> enough to do this without being a support burden will already know how.

SwartzCr:
fixes #213

### DIFF
--- a/_scripts/instruction-widget/data/inputs.json
+++ b/_scripts/instruction-widget/data/inputs.json
@@ -125,9 +125,21 @@
             "version": "0"
         },
         {
-            "name": "OpenBSD",
+            "name": "OpenBSD 6.0+",
             "id": "opbsd",
             "distro": "opbsd",
+            "version": "6"
+        },
+        {
+            "name": "OpenBSD 5.9"
+            "id": "opbsd",
+            "distro": "opbsd",
+            "version": "5.9"
+        },
+        {
+            "name": "OpenBSD (other)"
+            "id": "opbsd",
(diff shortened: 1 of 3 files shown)


SwartzCr:
@ohemorange wanna review?

ohemorange:
So keeping the references to `ports`?

SwartzCr:
okay! removed references to ports for openbsd

PR #222, opened 2017-05-04
add zesty as an option since it's now in the PPA

SwartzCr:


SwartzCr:
@ohemorange or @bmw wanna do a quick review?

PR #226, opened 2017-05-16
Make tests run in certbot-deploy

bmw:
Turns out our `certbot-deploy` script doesn't actually run tests. I was confused why #225 failed given it was exactly what I just had done a release on. In playing with it, I found the following output:
```
Running ["ScriptCheck", "ImageCheck", "LinkCheck"] on ./_site on *.html... 


Checking 0 external links...
Ran on 0 files!


HTML-Proofer finished successfully.
```
We need to build the website before we can test it. This is what's done in [Travis](https://github.com/certbot/website/blob/master/.travis.yml#L22) as well.

PR #228, opened 2017-05-19
Cache more in dockerfile

bmw:
Built on #226 so tests pass and to avoid merge conflicts. This PR simply moves the installation of `texlive` earlier in the Dockerfile, before we copy any local files that may have changed, to cache more between each build.


SwartzCr:
lgtm merge when you want

PR #231, opened 2017-05-29
Fix punctuation and formatting mistakes in privacy policy

zjs:
* Mark a command-line argument as code to avoid two hyphens becoming a dash.


zjs:
Rebased on #237; all that was left was the formatting of `--user-agent`.

PR #238, opened 2017-06-07
Cache more

### LINKED ISSUE #234: Move COPY .git* back to the bottom of the Dockerfile
After certbot/certbot#4770 lands, we can do this and cache more things making builds faster.

bmw:
Fixes #234.


bmw:
Apparently updating the submodule breaks all the things.

I'll deal with this later if no one else does.

PR #254, opened 2017-06-23
Updated the privacy policy to reflect an additional IRC service.

jgillula:
Updated the privacy policy to reflect an additional IRC service.


jgillula:
@SwartzCr Could you take a look and approve? Thanks!

bmw:
Great! We should also mention the change to the policy and provide a link to the previous version at the bottom. If someone makes that change, I'll review it and update the website.

jgillula:
Done #254

PR #255, opened 2017-06-23
add update info

SwartzCr:
change log for privacy policy


bmw:
Thanks for doing this before both Jeremy and I. Closing in favor of #258.

PR #256, opened 2017-06-23
update cron info with an example

SwartzCr:
Add links explaining cron and systemd
include a sample cron job


_scripts/instruction-widget/templates/getting-started/renewal.html
```diff
@@ -37,4 +32,12 @@
 revocation happened for some reason). Please select a random minute within the
 hour for your renewal tasks.</p>
 </aside>
-{{/cron_included}}
+
+<p>An example cron job might look like this, which will run at noon and midnight every day:</p>
+
+<pre>0 0,12 * * * {{base_command}} renew {{#cbauto}}--noself-upgrade{{/cbauto}}</pre>
```
bmw: I think by doing this there will be some fraction of users who just take this command as is. Additionally, we tell the user above to randomize the time it runs but we don't do it ourselves. I think we should change the command run to something like:
```
python -c 'import random; import time; time.sleep(random.random() * 60)' && {{base_command}} ...
```
I care less about this, but it might not hurt to pick a time other than 0 and 12 since that's what Debian and Ubuntu uses so the load on Let's Encrypt is further distributed. If you change this, don't forget to update the sentence about what this crontab does above.

Finally, the flag should be `--no-self-upgrade`.

_scripts/instruction-widget/templates/getting-started/renewal.html
```diff
@@ -18,16 +18,11 @@ <h3>Automating renewal</h3>
 </pre>
 
 If that appears to be working correctly, you can arrange for automatic renewal
-by adding a <tt>cron</tt> or <tt>systemd</tt> job which runs the
+by adding a <tt><a href="http://www.unixgeeks.org/security/newbie/unix/cron-1.html">cron</a></tt> or <tt><a href="https://en.wikipedia.org/wiki/Systemd">systemd</a></tt> job which runs the
```
bmw: Two things:

1. Can we change the text of this line to say "by adding a cron job or systemd timer which..." (with "systemd timer" being the link text)? This to me helps clarify what to look for if you go with systemd.
2. Can we use https://wiki.archlinux.org/index.php/Systemd/Timers for the link about systemd timers? I think it provides a pretty good distro agnostic description of the subject.

_scripts/instruction-widget/templates/getting-started/renewal.html
```diff
@@ -37,4 +32,12 @@ <h4>Note:</h4>
 revocation happened for some reason). Please select a random minute within the
 hour for your renewal tasks.</p>
 </aside>
-{{/cron_included}}
+
+<p>An example cron job might look like this, which will run at noon and midnight every day:</p>
+
+<pre>0 0,12 * * * python -c 'import random; import time; time.sleep(random.random() * 60)' && {{base_command}} renew {{#cbauto}}--no-self-upgrade{{/cbauto}}</pre>
```
bmw: This is my fault. I said we should use `random.random() * 60`, but this only spaces things out over a minute. We should do `random.random() * 3600` to spread it over the hour.

_scripts/instruction-widget/templates/getting-started/renewal.html
```diff
@@ -18,16 +18,11 @@ <h3>Automating renewal</h3>
...
+by adding a <tt><a href="http://www.unixgeeks.org/security/newbie/unix/cron-1.html">cron</a></tt> or <tt><a href="https://en.wikipedia.org/wiki/Systemd">systemd</a></tt> job which runs the
 following:<br>
 
 <pre>{{base_command}} renew {{#cbauto}}--no-self-upgrade{{/cbauto}}</pre>
 
-<p>More detailed information and options about renewal can be found in <a
-href="https://certbot.eff.org/docs/using.html#renewal">the full
-documentation</a>.</p>
-
-{{^cron_included}}
```
bmw: Rather than deleting these lines, I think we should actually expand their scope a bit. If cron is already configured on your system, we don't need to include text about how to set it up. I'd put this after the `--dry-run` command and put the closing tag after the example cron job.

_scripts/instruction-widget/templates/getting-started/renewal.html
```diff
@@ -16,20 +16,19 @@ <h3>Automating renewal</h3>
 <pre>
 $ sudo {{base_command}} renew --dry-run
 </pre>
+<p>
+More detailed information and options about renewal can be found in
+<a href="https://certbot.eff.org/docs/using.html#renewal">the full documentation</a>.
```
bmw: We're now linking to this section here and at the bottom of this file.

PR #262, opened 2017-07-06
Push files built by Travis to a build repo

### LINKED ISSUE #189: Automate website updates
We'd need to come up a convention on the Certbot side to prevent changes about unreleased features from going live, but this saves devs time and causes website improvements to be deployed faster.

vbrown608:
This resolves #189 on the Travis/Github end. Following Noah's original suggestion, we'll still need to implement step (4) where the person doing the deploy pulls the built files to an internal repo from which certbot.eff.org pulls on cron.

(This PR is also a test that the right thing happens when there's a PR :+1: )


vbrown608:
Okay, I think the right thing is happening on PRs. This is ready for review.

Here's the expected behavior:
- If the analogous branch in website-builds doesn't exist yet, it's cut from master (to improve diffs against master).
- The build is committed to the branch.
- PR builds are committed to the analog of the source branch, even though they simulate a merge into master. I think this will keep the website-builds repo cleanest. They're marked with a special commit message that has the PR number in it.

I set Travis to build all branches so folks can see their build without needing to open a PR. If that's creating too much noise we can go back to just building PRs and branches prefixed with `test-`.

push-build.sh
```diff
@@ -0,0 +1,38 @@
...
+git fetch upstream master
+git branch $BRANCH upstream/master
+# Manually set which branch HEAD points to to avoid changes to the
+# working tree caused by "git checkout".
+git symbolic-ref HEAD refs/heads/$BRANCH
+git reset
+
+# If this branch already exists upstream, use that instead.
+git fetch upstream $BRANCH
+git reset upstream/$BRANCH
```
bmw: Perhaps my git fu isn't strong enough, but is there any difference in what's pushed to an existing branch using this approach rather than just using that branch above instead of `master`? Does the difference matter? If we changed the approach we'd have to check for the existence of the branch in the website-builds repo.

If this works cleanly as is though I'm fine with it.

vbrown608:
https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions

Looks like the key will only be decrypted for PRs that originate from the same repo - not for PRs across across forks. This seems like a significant barrier to the usefulness of this script and I'm not sure if there's a way to avoid it :/ Even deploy methods that are built into Travis (ex [GitHub releases](https://docs.travis-ci.com/user/deployment/releases/)) require decrypting a secret and making it available to .travis.yml. I'm not sure there's a way to allow Travis to do something with the built files during a PR without allowing anyone to do that thing.

bmw:
Ah that's unfortunate but makes sense.

How about we set this up to run on all branches but exclude PRs using an `if` statement in our `.travis.yml` as shown in your link? This will keep a built version of `master` up to date allowing us to easily deploy the website and anyone reviewing PRs that wants to see the files built automatically can simply create a branch on this repo containing the PR.

.travis.yml
```diff
@@ -22,15 +24,19 @@ script:
   - gulp build
   - npm test
 
+after_success:
+  - chmod +x push-build.sh
```
vbrown608: :+1:

.travis.yml
```diff
@@ -22,15 +24,19 @@ script:
...
+    # speeds up installation of html-proofer
+    - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
     - NODE_ENV=production
 
 # Only build pushes to the master branch, PRs, and branches beginning with `test-`.
 branches:
   only:
-    - master
-    - /^test-.*$/
+    - /.*/
```
vbrown608: My thinking was that users who push to the repo might want to preview their build without opening a PR. Now that we won't be building PRs we'll need this.

Good point about the section and thanks for catching that comment!

push-build.sh
vbrown608: Right now, lines 31 and 32 fail if upstream/$BRANCH doesn't exist, leaving us with with new branch off master. We need to fetch upstream/$BRANCH if it exists or the feature branch will be in conflict with upstream when we try to push.

I rewrote this to use a conditional rather than a sometimes-expected fail, which I hope makes things a little clearer. Does that address your concern, or am I misunderstanding?

README.md
```diff
@@ -66,3 +66,11 @@ To run the tests:
 npm test
 ```
 (Files with known issues are ignored.)
+
+## Travis Builds
+All branches and pull requests and built and tested by Travis. The built files are pushed to an analagous branch in [certbot/website-builds](https://github.com/certbot/website-builds).
```
bmw: nit: Pull requests aren't sent to website builds.

.travis.yml
```diff
@@ -9,6 +9,8 @@ addons:
       - texlive-latex-extra
 
 before_install:
+  - openssl aes-256-cbc -K $encrypted_e5009b71da55_key -iv $encrypted_e5009b71da55_iv
```
bmw: I think we need to conditionally do this if it's not a PR as well to prevent https://travis-ci.org/certbot/website/builds/253435308#L371.

PR #263, opened 2017-07-07
Use exact package names for CentOS/RHEL

davidjb:
These packages are installable with the `python2-` prefix.  I'm unsure whether the packages changed name and can't seem to find why the discrepency, but EPEL is currently using these names:  see https://mirror.aarnet.edu.au/pub/epel/7/x86_64/p/.

### DIFF
--- a/_scripts/instruction-widget/install.js
+++ b/_scripts/instruction-widget/install.js
@@ -86,9 +86,9 @@ module.exports = function(context) {
       context.packaged = true
 
       if (context.webserver == "apache") {
-        context.package = "certbot-apache";
+        context.package = "python2-certbot-apache";
       } else if (context.webserver == "nginx") {
-        context.package = "certbot-nginx";
+        context.package = "python2-certbot-nginx";
       }
     }
   }


bmw:
My understanding is that these packages are also installable without the `python2` prefix. I've tested that this works on CentOS 7.

@hogarthj, what should we be instructing people to do here?

davidjb:
For what it's worth, the name difference confused me and I didn't proceed with installing until I investigated.  If the plan is to leave the instructions as is, I'd suggest adding a note to the install instructions explaining the difference.

SwartzCr:
@davidjb would you be willing to write up such a note?

davidjb:
@SwartzCr [Fedora's package naming conventions](https://fedoraproject.org/wiki/Packaging:Naming#Python2_binary_package_naming) (which [EPEL follows](https://fedoraproject.org/wiki/EPEL/GuidelinesAndPolicies)) require Python 2 binary packages to be prefixed with `python2-` so isn't it better to specify the exact/correct package name in the first instance?

tchollingsworth:
RHEL 7 uses Python 2 but RHEL 8 will probably use Python 3. Why make ourselves update the documentation and everybody update their scripts in a couple years when we can just suggest something that is future-proof?

davidjb:
The unexpected difference in name made me pause installation and dig into why a package with the exact name wasn't being installed. My thought pattern at the time was that these various Python _libraries_  were being installed (e.g. All of the dependencies simultaneously installed start with python2- as well), and where was the package with the executable itself? After having gone and looked up EPEL's naming convention, it now makes sense but that's why I raised the issue. 

Perhaps others accept whatever packages they're been told to install without reviewing what they are. That's a touch concerning if that's a practice, but in my case I was investigating as if something was wrong. Hence the PR to alleviate similar confusion. 

In terms of maintainability, major RHEL releases don't come around that often plus neither do shifts in major Python version.  The code changes required for package name changes are minor and changes in a future major OS version probably mean other docs updates would be required anyway.

tchollingsworth:
I just submitted pull request #281 which would add the following note:

> This command will install the certbot packages corresponding to the system default Python interpreter, so the packages that are actually installed will contain a prefix like `python2-`.

Would this satisfy your concerns? I just hate teaching everyone a command we know is going to break in a couple years when we could just teach people about package aliases instead. 😀

davidjb:
@tchollingsworth Definitely better as it gives an explanation, but I held back suggesting that because the wording has to be vague.  

Bear in mind that `yum` will eventually be replaced with `dnf` in a future EL version (as it has already been in Fedora) so the instructions will already need reworking at that point (eg when Python 2 goes to Python 3).

hogarthj:
Just a quick note... the last I heard was that yum was not going to change
to dnf.

Instead some of the last things that were missing from dnf that used to be
in yum were being added to yum and although the same backend would exist on
RHEL and Fedora the actual command I  RHEL8 is still planned to be yum.

This is going off the RHEL yum maintenance team on the fedora development
lists a few months back.

PR #267, opened 2017-07-07
Add Devuan instructions

bmw:
Add instructions for Devuan based on the conversation in certbot/certbot#4903.

### DIFF
--- a/_scripts/instruction-widget/install.js
+++ b/_scripts/instruction-widget/install.js
@@ -95,6 +95,7 @@ module.exports = function(context) {
 
   debian_install = function() {
     template = "debian";
+    context.devuan = context.distro == "devuan"
 
     // Debian Jessie
     context.base_command = "certbot";
@@ -107,8 +108,8 @@ module.exports = function(context) {
       context.certonly = true;
     }
 
-    // Debian Jessie backports.
-    if (context.version == 8) {
+    // Jessie backports.
+    if ((context.devuan && context.version == 1) || context.version == 8) {
       context.backports_flag = "-t jessie-backports";
     }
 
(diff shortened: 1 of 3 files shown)


maurorappa:
Hi, any ETA on this? :)

_scripts/instruction-widget/install.js
```diff
@@ -95,6 +97,7 @@ module.exports = function(context) {
 
   debian_install = function() {
     template = "debian";
+    context.devuan = context.distro == "devuan"
```
SwartzCr: Won't this make every debian install devuan? Since it's in the body of the debian_install function the context.devuan flag will be truethy meaning that everyone will get the devuan backports instructions

_scripts/instruction-widget/install.js
```diff
@@ -107,8 +110,8 @@ module.exports = function(context) {
       context.certonly = true;
     }
 
-    // Debian Jessie backports.
-    if (context.version == 8) {
+    // Jessie backports.
+    if ((context.devuan && context.version == 1) || context.version == 8) {
```
SwartzCr: Is it true that on devuan 0 you'll not need to enable backports, but will be able to use apt to install it?

SwartzCr:
Sorry for the extreme delay!
Just left some comments for brad - hope to get this merged soon

_scripts/instruction-widget/install.js
bmw: Yes. Devuan is very similar to Debian. If you have Devuan testing or unstable, it's available in the official repos.

_scripts/instruction-widget/install.js
bmw: `context.devuan` is set to true only if `context.distro` is "devuan"`. If it is, we'll use the Devuan instructions, otherwise, we'll use the Debian instructions.

SwartzCr:
okay - lgtm, merging!

PR #275, opened 2017-10-02
Make the right margin wider in the docs

pde:
* This makes the table at
   https://certbot.eff.org/docs/using.html#getting-certificates-and-choosing-plugins
   much more readable on many devices

### DIFF
--- a/_sass/theme.css
+++ b/_sass/theme.css
@@ -3715,6 +3715,7 @@ p {
   margin: 0;
   font-size: 16px;
   margin-bottom: 24px;
+  max-width: 800px;
 }
 
 h1 {
--- a/_sass/theme.css
+++ b/_sass/theme.css
@@ -4561,7 +4561,6 @@ div[class^='highlight'] pre {
 .wy-nav-content {
   padding: 1.618em 3.236em;
   height: 100%;
-  max-width: 800px;
   margin: auto;
 }
 


maximillianh:
All of the literature I've seen puts the optimal line length for reading speed and comprehension between 50-80 characters.  Removing the max-width increases characters per line to 200+.

Reformatting for a stacked table on small screens may be a better solution, i.e. [](https://sitesforprofit.com/wp-content/uploads/2014/05/resptable.jpeg)

pde:
Here's what this looks like for me at present:

tables with scrollbars despite lots of horizontal space

It seems crazy to have all that space that's unused, and then scrollbars on the table. Perhaps we can remove the margins and then do something else to limit the width of the `<p>` elements?

bmw:
The unused space feels counterintuitive to me too, but I trust Max on this and this PR hasn't seen any progress in a while so I'm going close it.

pde:
Augh, that nerd-sniped me. 

I updated the PR so it widens the tables (and other useful things like command lines), but leaves the text narrow. Unfortunately, this looks a bit weird and will need some more manual attention from someone better than me at CSS to make it all look pretty in this more legible form.

pde:
(Reopen or not, as you wish)

bmw:
Heh sorry. Wasn't trying to nerd-snipe you. Just trying to clean out the PR queue.

@maximillianh, is there someone from E&D who can provide feedback on this?

pde:
Further things this PR would need to be in great shape:

* [ ] Style the preformatted blocks of code so that they aren't wider than their contents (but can be arbitrarily wide when needed)
* [ ] Style the tables so that they aren't wider than their total contents
* [ ] Limit the "warning" and "note" blocks to 800px to line up with the text

PR #281, opened 2017-10-06
add note about package aliases

tchollingsworth:
This satisfies the concerns in #263 without sacrificing forward-compatibility and
keeping things shorter to type and easier to remember.

The partial is added to Fedora and CentOS/RHEL where it currently applies, but
could be added to other distros where it is relevant.

The package name prefix used by each distro is defined in
{{python_prefix}}, and the note can be switched off by making that falsy. This
is done when just certbot is installed, since that isn't an alias.


bmw:
@tchollingsworth, while I certainly appreciate you trying to save us maintenance burden with Python changes, I went ahead and merged #263 to use the `python2-` package names for RHEL 7 based systems. My main motivating factor here is certain commands like `yum info certbot-apache` don't work and you have to use the full name of the library.

Because that was merged, I'm going to go ahead and close this PR. Sorry this sat around so long!

Issue #283, opened 2017-10-16
Update Debian backports instructions


bmw:
Stretch users should use `stretch-backports` and Jessie users should use `jessie-backports-sloppy`.

bmw:
We should make sure that packages are in these repos before we push these changes to the live site.

SwartzCr:
lol - It would have been nice to have been mentioned in this comment so that I would have noticed to do this before @mrphs emailed me about it. Working on a PR now

bmw:
I mentioned this issue to you in #285 and also thought I brought it up verbally a while ago, but I'm sorry you weren't as aware of the problem as you'd have liked. If in general there are categories of issues you'd like me to mention you on or things I can do to better communicate the state of the project to you, let me know.

It's worth noting that a newer version of Certbot is not yet in `jessie-backports-sloppy`. I brought this up to Harlan last week and he's going to work on changing that.

SwartzCr:
Okay I made a commit, that at least fixes it for stretch, and we can fix it again when we're in jessie sloppy

alexzorin:
Any news on getting into `jessie-backports-sloppy`?

If Certbot is not going to land there, perhaps the instructions could be updated to use `certbot-auto`, as the instructed `jessie-backports` is giving an ancient version.

PR #285, opened 2017-11-09
suggest nginx package

SwartzCr:
suggest nginx package to debian users: https://packages.debian.org/buster/python-certbot-nginx

### DIFF
--- a/_scripts/instruction-widget/install.js
+++ b/_scripts/instruction-widget/install.js
@@ -107,7 +107,7 @@ module.exports = function(context) {
     if (context.webserver == "apache") {
       context.package = "python-certbot-apache";
     } else if (context.webserver == "nginx") {
-      context.certonly = true;
+      context.package = "python-certbot-nginx";
     }
 
     // Jessie backports.


SwartzCr:
According to this search: https://packages.debian.org/search?searchon=names&keywords=python-certbot-nginx
It's available in stretch and stretch backports
The version native in stretch is the same version as the certbot package (10.2-1): https://packages.debian.org/stretch/certbot
It is also the same version as the certbot in jessie backports: https://packages.debian.org/jessie-backports/python-certbot-nginx
Arguably stretch and jessie should be bumped in some way to 19, but for now this seems identical to our existing documentation

bmw:
Erica and I decided [here](https://github.com/certbot/certbot/issues/4079#issuecomment-296333774) when sid had 0.11.1 and every other Debian release (that had an Nginx package) had 0.10.2 to wait until they were updated before telling people to use them.

If you want to talk about this again we can, but the Nginx plugin was in a pretty rough state at this point. I'd personally rather see #283 be resolved and tell people to install the packages from backports.

PR #288, opened 2017-12-12
suggest stretch backports and nginx

SwartzCr:
addresses #283 until 0.19.0 is in stretch sloppy
this will give stretch users access to the nginx plugin: https://packages.debian.org/jessie-backports/python-certbot-nginx

### DIFF
--- a/_scripts/instruction-widget/install.js
+++ b/_scripts/instruction-widget/install.js
@@ -123,6 +123,9 @@ module.exports = function(context) {
         context.package = "python-certbot-nginx";
       }
     }
+    if (context.version == 10 && context.webserver == "nginx") {
+      context.package = "python-certbot-nginx";
+    }
 
   }
 
--- a/_scripts/instruction-widget/install.js
+++ b/_scripts/instruction-widget/install.js
@@ -108,16 +108,20 @@ module.exports = function(context) {
 
     if (context.webserver == "apache") {
       context.package = "python-certbot-apache";
-    } else if (context.webserver == "nginx") {
-      context.package = "python-certbot-nginx";
     }
 
     // Jessie backports.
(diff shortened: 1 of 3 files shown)


bmw:
I think tests should pass if you merge `master` into this PR.

_scripts/instruction-widget/install.js
```diff
@@ -98,6 +98,8 @@ module.exports = function(context) {
   debian_install = function() {
     template = "debian";
     context.devuan = context.distro == "devuan"
+    context.jessie = context.version == 8
+    context.stretch = context.version == 9
```
bmw: It looks like `context.stretch` is being set to true for testing/unstable too causing it to tell you to install Stretch backports on that system too. This is due to: https://github.com/certbot/website/blob/3c8502138130f49a10bd95b6d33aaa2a51782cf3/_scripts/instruction-widget/data/inputs.json#L53 We should update this to a new value to avoid this problem now and in the future.

_scripts/instruction-widget/templates/install/debian.html
```diff
@@ -12,7 +12,7 @@
 {{/devuan}}
 {{^devuan}}
 First you'll have to follow the instructions <a
-href='http://backports.debian.org/Instructions/'>here</a> to enable the Jessie
+    href='http://backports.debian.org/Instructions/'>here</a> to enable the {{#jessie}}Jessie{{/jessie}}{{#stretch}}Stretch{{/stretch}}
```
bmw: This indentation looks strange to me. Did you do it on purpose to show that this is part of the  `<a>` tag from the previous line? If so, feel free to leave it as is, I just wanted to make sure it was intentional.

PR #290, opened 2017-12-13
Smooth scroll to instructions after selecting os and server

### LINKED ISSUE #111: Smooth scroll to instructions
After a use selects their OS and webserver, animate scroll to instructions so they don't get disoriented.

vbrown608:
* Remove javascript error by taking out old wink code
* Smooth scroll to instructions. I think this will be less disorienting for users.
Closes #111.


vbrown608:
This is now part of #295.

Issue #294, opened 2017-12-20
Deprecate all-instructions page?


vbrown608:
certbot.eff.org/all-instructions was originally created to support users without javascript. As part of #261 I'm making the main instruction widget at certbot.eff.org support no-js users, so that page is no longer needed. 

Should we
(a) Redirect that page to certbot.eff.org or
(b) Keep the list of links, but link to the standalone instruction page for each os, webserver pair (these will now be accessible to no-js users).

According to anon-stats the all-instructions page gets about 80 page views per day, or 0.8% of our traffic. (a) would be nice if we can get away with it (less cruft from the past) but I'm fine with either.

bmw:
I'm fine with (a), but it might be good to some more opinions.

cc @jsha, @SwartzCr

jsha:
I'm in favor of (a)

SwartzCr:
I'm fine with option a

PR #295, opened 2018-01-03
Render full instruction pages at build time

### LINKED ISSUE #261: Improve search engine findability for help pages
The current one-page design is nice and fast, but it's hard for search engines to find. For instance, the instructions to install on Ubuntu are at https://certbot.eff.org/#ubuntutyakkety-apache. Because the locator for that specific page is based on a URL fragment instead of the whole URL, search engines see it as identical to https://certbot.eff.org/.

We should break out each platform/software combination into its own page, with a full URL, like https://certbot.eff.org/lets-encrypt-ubuntu-yakkety-apache. Note that the # is removed, and the keywords people are likely to search for are separated out with hyphens. We should also ensure each page has a title that reflects what the page is about, e.g. "How to get a Let's Encrypt certificate on Ubuntu with Certbot and Apache," and a 1-sentence blurb at the top to give context to people arriving directly from a search rather than from the main page.

vbrown608:
Renders the instruction page for each (os, webserver) pair at build time. This lets us create standalone pages that are accessible to users with javascript disabled. It also moves more rendering and routing into jekyll/nginx, which I think is preferable to the very custom javascript.

When users with javascript enabled submit the widget form, I'm using js to update the url + page history to the new path. For users without javascript, the form submits to something like certbot.eff.org/?server=apache&os=arch. Nginx parses the query string and rewrites it to the correct url.

Individual instruction pages are now found at, eg, certbot.eff.org/i/arch-apache. Do we want to tweak that to include a keyword about letsencrypt? We could do certbot.eff.org/letsencrypt-instructions/arch-apache, for example.

We'll need to update the nginx config before we deploy this. I'm working with techops on that.

Resolves #261, #294 and #111


README.md
```diff
@@ -4,9 +4,17 @@ Website for [EFF's Certbot](https://certbot.eff.org/) project. Uses Jekyll for s
 
 [![Build Status](https://travis-ci.org/certbot/website.svg?branch=master)](https://travis-ci.org/certbot/website)
 
-## Getting Started
+## Development
 
-### Install
+### With Docker
+
+`docker-compose up`
```
wioux: The docker build fails if you don't have the _docs submodule cloned. Those steps (4/5) are now under "Without Docker" which is a little confusing.

all-instructions/index.html
```diff
@@ -1,53 +0,0 @@
----
```
wioux: The docker build fails when it tries to copy this deleted directory https://github.com/certbot/website/blob/f1c9566c5b60f7e71d78c631fe19916f64a5b79d/Dockerfile#L53

docker-compose.yml
```diff
@@ -0,0 +1,17 @@
...
+    image: nginx
+    ports:
+      - '4000:4000'
+    volumes:
+      - assets:/usr/share/nginx/html:ro
+      - ./nginx.conf:/etc/nginx/conf.d/default.conf
+  site:
+    build: .
+    volumes:
+      - .:/opt/certbot/
```
wioux: This volume caused some issues for me. `npm install` / `_docs.sh depend/install` / `gulp build`  are run in the container as part of the Dockerfile, so my host directory did not initially have node_modules / build products.

_scripts/instruction-widget/main.js
```diff
@@ -90,6 +81,15 @@ InstructionWidget = (function() {
     }
   };
 
+  // Users used to be able to link to an instruction set with an anchor link.
+  // We can redirect them to a standalone page.
+  redirect_anchor = function() {
+    var params = window.location.hash.replace('#', '').split('-');
+    if (params.length === 2) {
+      jump(params[0], params[1]);
```
wioux: This rewrote the url for me but didn't render any instructions.

http://localhost:4000/#debianjessie-nginx

all-instructions/index.html
vbrown608: Thanks. I'll make a follow-up PR to handle this with docker-ignore. I didn't want to jam too much (more) into this one.

wioux:
Everything looks good. :+1:

I think what jsha suggested about page titles in #261 is a good idea worth pulling out into a new issue.

vbrown608:
Thanks @wioux, I'll merge this when changes to the nginx config are live.

Issue #299, opened 2018-01-18
Add Ubuntu Artful


jsoref:
Ubuntu Zesty has reached End Of Life ->
https://certbot.eff.org/all-instructions/#ubuntu-17-04-zesty-apache

Ubuntu Artful should be listed.

```certbot-ubuntu-certbot-artful.list
deb http://ppa.launchpad.net/certbot/certbot/ubuntu artful main
# deb-src http://ppa.launchpad.net/certbot/certbot/ubuntu artful main
```

SwartzCr:
@jsoref any chance you want to make a PR for this?

jsoref:
Yes, although not immediately

bmw:
Fixed in #308.

PR #308, opened 2018-01-25
Fix up PPAs

jsoref:
This covers #299

### DIFF
--- a/_data/inputs.json
+++ b/_data/inputs.json
@@ -59,6 +59,12 @@
             "version": "0"
         },
         {
+            "name": "Ubuntu 17.10 (artful)",
+            "id": "ubuntuartful",
+            "distro": "ubuntu",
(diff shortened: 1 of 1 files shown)


bmw:
Sorry that I'm just getting around to this as Artful approaches it's EOL, but if you resolve merge conflicts, I'll merge this PR.

bmw:
As mentioned in IRC, our PPA is actually behind bionic right now so we should just tell people to use the packages from there for now. To do this,`_scripts/instruction-widget/install.js` and maybe `_scripts/instruction-widget/templates/install/ubuntu.html` should be modified.

Bonus points for doing this in a way we can easily revert because this should only be temporary. Let me know if you're not able to pick this up.

bmw:
Opened #341 to track Bionic instructions.

PR #309, opened 2018-01-25
ppa: drop yakkety + zesty as they hit EOL

jsoref:
Note that these also had a c+p error (the `t` appears to come from `ubuntutrusty`), but as they're EOL, there's no reason to maintain that...

### DIFF
--- a/_data/inputs.json
+++ b/_data/inputs.json
@@ -59,18 +59,6 @@
             "version": "0"
         },
         {
-            "name": "Ubuntu 17.04 (zesty)",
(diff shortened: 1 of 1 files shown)


jsoref:
This addresses the other half of #299.
I'm making them as distinct commits because I can understand someone saying that the instructions should remain even though it is at EOL.

bmw:
Sorry for the delay here @jsoref.

If you can resolve the merge conflicts, I'll merge this PR.

PR #314, opened 2018-02-13
Link to /docs/ instead of /docs

th0114nd:
https://certbot.eff.org/docs has approximately the same text, but is missing the CSS and every link 404s.
https://certbot.eff.org/docs/ loads the CSS and links work.


vbrown608:
This is related to #315. I think it's a good temporary improvement.

bmw:
Thanks for the PR @th0114nd. This was fixed in #317.

Issue #315, opened 2018-02-15
Read The Docs should not use relative paths for assets


vbrown608:
Read The Docs builds the theme for the documentation using relative paths, ex `<link rel="stylesheet" href="_static/css/theme.css" type="text/css" />`. This causes these assets to 404 when the browser loads certbot.eff.org/docs instead of certbot.eff.org/docs/. We should see if we can configure Read The Docs to output the correct absolute path.

PR #316, opened 2018-02-16
Render full instruction pages for SEO, all-instructions page for no-js users

vbrown608:
This PR adds back backend-rendered standalone instruction pages for better SEO. 

That functionality was removed because of problems with the redirection from the query string generated by the form the standalone pages. Instead, we'll support users with Javascript disabled the old way - by generated a page that lists every instruction set.

This PR also redirects /docs to /docs/. Static assets don't load correctly on the /docs page for the reason described in #315. I think landing on /docs a side effect of the try_files directive in the updated nginx config.

The current nginx config supports the changes in this PR, so this can be deployed as soon as it's approved.


_scripts/main.js
```diff
@@ -26,6 +26,11 @@ var Raven = require('raven-js');
 })(Raven);
 
 (function() {
+  [].forEach.call(document.querySelectorAll('.js-only'), function (el) {
+    console.log("setting visibility");
```
wioux: Looks like a debugging statement got committed.

nginx.conf
```diff
@@ -9,4 +8,7 @@ server {
         try_files $uri/index.html $uri.html $uri/ $uri =404;
     }
 
+    location = /docs {
+        return 301 /docs/ ;
```
wioux: This redirect ditches the URL's port component, so I get sent from `localhost:4000/docs` to `localhost/docs` where there is no server running.

Separately, if I go directly to `http://localhost:4000/docs/` it seems there's nothing there? I get a 404.

vbrown608: Thanks for catching this. I talked to Starchy about it today - he will add this rule to the load balancer rather than the nginx container. Which also means at this point this nginx.conf is really just a convenience for building the site with Docker and not a a mirror of the production container.

For the development version, I went ahead an made the ports match inside the container and outside the container again.

nginx.conf
```diff
@@ -1,4 +1,5 @@
 server {
+    listen: 4000
```
wioux: The nginx container crashes here. Looks like we need to ditch the colon and add a semicolon.

vbrown608: Ack sorry about that.

wioux:
:rocket: :moon:

PR #332, opened 2018-04-13
329 warning banner

lschatzkin:
I spoke to @pde and he felt the warning just need to be on front page.

### DIFF
--- a/_includes/warning.html
+++ b/_includes/warning.html
@@ -1,0 +1,7 @@
+<div>
+  <aside class="warning" id="wc-warning">
+    <p>
+      Trying to get a wildcard certificate?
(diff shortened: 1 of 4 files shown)


_includes/warning.html
```diff
@@ -0,0 +1,7 @@
+<div>
+  <aside class="warning" id="wc-warning">
+    <p>
+      Trying to get a wildcard certificate?
+      Please use the dropdown menus below to get instructions specific to your system, and read those instructions carefully.
```
bmw: Are we missing the closing `</p>` tag here or was this intentional?

lschatzkin: Not intentional, thanks for catching that!

PR #337, opened 2018-05-21
Don't push if an error occurs.

bmw:
If an error occurs in our build script, exit immediately rather than pushing to
website-builds. The check if "$BRANCH" exists had to be rewritten slightly to
accommodate this.


vbrown608:
Looks good to me!

Issue #323, opened 2018-03-15
Redirect fedora24 id to fedora


bmw:
Once upon a time, we had different instructions for different Fedora versions. Back then, we used the id `fedora24` to refer to Fedora 24+ instructions where Certbot was packaged. Fedora 24 is now EOL'd and the oldest version of Fedora that's still supported is Fedora 26.

I'd like to change the id used for Fedora in the instruction generator to just be `fedora`, but I don't want to break any links people have to the site since the `fedora24` instructions are identical to the instructions for all Fedora versions now: Get Certbot from your OS package manager.

@vbrown608, do you have any ideas on the best way to accomplish this?

vbrown608:
@starchy can we do this with nginx?

starchy:
Is this just a URL rewrite or is there anything else required?

vbrown608:
Yep, just a rewrite.

starchy:
@bmw since I think we're talking about an EFF infrastructure change, not a code change, do you want to open a ticket in that bug tracker listing the URL mappings?

bmw:
Sure I'll open a ticket.

@vbrown608, to make this a smooth transition, I think it'd be good if we served the instructions for both the `fedora` and `fedora24` ids. Is there an easy way to do this without adding a 2nd Fedora option in the drop down menu?

PR #344, opened 2018-06-12
Add generic fedora OS and hide fedora24

vbrown608:
Part of #323

This creates identical pages at `/lets-encrypt/fedora-apache` and `/lets-encrypt/fedora24-apache`, but excludes `fedora24` from the dropdown options.

I think an nginx redirect would still ultimately be cleaner, but this will help with the switchover.

### DIFF
--- a/_data/inputs.json
+++ b/_data/inputs.json
@@ -101,6 +101,12 @@
             "version": "26"
         },
         {
+            "name": "Fedora 26+",
+            "id": "fedora",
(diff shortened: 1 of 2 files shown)


vbrown608:
(I think this is failing due to an unrelated issue - we check that all external links in the site work, and https://devuan.org/os/etc/apt/sources.list.d/devuan-stable-backports.list is 404ing)

PR #345, opened 2018-06-15
Fix devuan link

bmw:
@vbrown608, this should fix test failures in #344. If you're able to review it, that'd be great, otherwise, let me know and I can try to wrangle someone from tech projects to review.

### DIFF
--- a/_scripts/instruction-widget/templates/install/debian.html
+++ b/_scripts/instruction-widget/templates/install/debian.html
@@ -6,8 +6,8 @@
 {{#backports_flag}}
 <p>
 {{#devuan}}
-First you'll have to enable the Jessie backports repo by enabling backports in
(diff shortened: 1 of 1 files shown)


vbrown608:
Looks good! BTW we can remove that check if it's not useful and just check internal links.

bmw:
Thanks for the review!

Let me know if you disagree, but I personally think it's useful so we can fix links when the site they link to changes, however, it has the downsides of causing unrelated failures in the tests for PRs.

How about I set up Travis to run tests on master every 24 hours and send a message to Certbot's IRC channel when the tests fail? Hopefully that allows us to catch these things sooner and clean them up before they affect incoming PRs.

vbrown608:
Sounds good to me

PR #371, opened 2018-10-26
Ensure Debian/Devuan installations have HTTP-01 support

bmw:
The changes made in this PR were:

* Tell Debian & Devuan Jessie users to use `certbot-auto` rather than the old version of Certbot without HTTP-01 support that's available in their backports repo.
* Add instructions for Devuan ASCII (which for our purposes is basically the same as Debian Stretch)

### DIFF
--- a/_data/inputs.json
+++ b/_data/inputs.json
@@ -149,16 +149,22 @@
             "version": "10"
         },
         {
+            "name": "Devuan ASCII 2.0",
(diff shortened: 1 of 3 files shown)


bmw:
Thanks jsha!

If anyone else wants to look this over feel free, but I feel confident about the changes I made here.

PR #379, opened 2018-11-20
Node and gulp-sass upgrade

### LINKED ISSUE #378: Builds are failing
Website builds are failing when running `gulp build` due to a `TypeError`. See https://travis-ci.org/certbot/website/builds/456766639#L1094.

@vbrown608, are you able to take a look at this?

vbrown608:
Fixes #378

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - 4.4.0
+  - 8.12.0
 
 addons:
   apt:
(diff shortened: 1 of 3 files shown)


bmw:
Thanks for the quick turnaround on this @vbrown608!

Does https://github.com/certbot/website/blob/master/README.md#install need to be updated? Do we need node 8+ now?

vbrown608:
Thanks for catching that! I think we just need 6 or higher at the moment but let's put 8 in the README so it matches Docker and Travis. 8 is still a couple versions behind and other things may break for 6.

PR #385, opened 2018-11-29
Upgrade packages and track package-lock.json

vbrown608:
- Track package versions
- Upgrade some packages
  - In particular, upgrade from Gulp 3 to Gulp 4, which required some code changes


_gulp/index.js
```diff
@@ -6,6 +6,11 @@ tasks.forEach(function(task) {
   require('./tasks/' + task);
 });
```
mfb: So what's the rationale on various task definitions changing?

vbrown608: Sorry, a comment in the PR would have been helpful for the change below, which isn't really part of the upgrade. `instructions`, `css`, and `js` used to be prerequisites for `jekyll:watch` and `jekyll:build` - this makes them peers of those tasks, which all come together to form the `watch` task. They can be run independently of building/watching Jekyll and I think this represents that situation better.

A few other comments about the upgrade:
1. In the previous syntax, we could only pass a list of tasks that would run before the main text (eg below, `watch` would run once `jekyll:watch` and `serve` completed). Now, we use the `parallel` and `series` fns to describe build order. Doing so gives us more flexibility so we can build up tasks like `docs:install` in `_gulp/tasks/docs.js`.

2. Upgrade docs I read recommended moving away from anonymous functions, so I've done that throughout the tasks. From https://codeburst.io/switching-to-gulp-4-0-271ae63530c0:
> If you’re switching to gulp@4.0, and you haven’t already, you’ll likely want to stop using anonymous functions. They print \<anonymous\> when run and aren’t very useful when viewing task metadata.

PR #387, opened 2019-01-08
Update DNS plugin install guide to specify package prefixes per OS

### LINKED ISSUE #362: Be more specific about DNS plugin package names
https://github.com/certbot/website/blob/5b9245babe160d7f8e2fae728d2a824bb39e8bc7/_scripts/instruction-widget/templates/install/dnsplugins.html#L8-L10

It's not imminently clear what the name of the DNS plugin package is without performing an `apt-cache search` or `yum search`. For instance, depending on how it's packaged, it could be `python2-certbot-dns-*` or `python3-certbot-dns-*`.

sydneyli:
fixes #362 
 * Specifies `dns_package_prefix` for OSes where DNS plugin is packaged.
 * Rewords DNS plugin install text to make it more clear whether a particular OS has it packaged, and to surface the DNS package name.

reviewer should double-check that these package names are correct

### DIFF
--- a/_scripts/instruction-widget/install.js
+++ b/_scripts/instruction-widget/install.js
@@ -23,6 +23,7 @@ module.exports = function(context) {
 
     context.cron_included = false;
     context.dns_plugins = false;
+    context.dns_package_prefix = "";
     context.jessie = false;  // Special jessie instructions for certbot-auto
     // Each case listed here should map to a template.
     // They don't necessarily need to map to distros.
@@ -101,6 +102,7 @@ module.exports = function(context) {
         context.package = "python2-certbot-nginx";
       }
       context.dns_plugins = true;
+      context.dns_package_prefix = "python2-certbot-dns"
     }
   }
 
@@ -109,6 +111,7 @@ module.exports = function(context) {
     context.devuan = context.distro == "devuan"
 
     context.dns_plugins = true;
+    context.dns_package_prefix = "python3-certbot-dns"
 
     context.base_command = "certbot";
     context.cron_included = true;
@@ -142,6 +145,7 @@ module.exports = function(context) {
     context.base_command = "certbot";
     context.cron_included = true;
     context.dns_plugins = true;
+    context.dns_package_prefix = "python3-certbot-dns"
   }
 
   gentoo_install = function() {
@@ -170,6 +174,7 @@ module.exports = function(context) {
     context.base_command = "certbot";
     context.base_package = "certbot";
     context.dns_plugins = true;
+    context.dns_package_prefix = "certbot-dns";
   }
 
   fedora_install = function() {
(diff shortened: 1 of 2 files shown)


_scripts/instruction-widget/install.js
```diff
@@ -183,6 +188,7 @@ module.exports = function(context) {
       context.package = "certbot-nginx";
     }
     context.dns_plugins = true;
+    context.dns_package_prefix = "certbot-dns";
```
bmw: This should be `python3-certbot-dns`.

(It appears the aliases without the `python-` prefix weren't added for our DNS plugins. I talked to our Fedora maintainer who said they'd add them, but they said that we should tell people to use `python3-*` in the meantime.)

_scripts/instruction-widget/templates/install/dnsplugins.html
```diff
@@ -1,21 +1,23 @@
...
-name of the DNS plugin you want to install.
+Certbot's DNS plugins <strong>are available</strong> for your system. These
+plugins can be used to automate obtaining a wildcard certificate from Let's
+Encrypt's ACMEv2 server.  To use one of these plugins, you must have configured
+DNS for the domain you want to obtain a certificate for with a DNS provider
+that Certbot has a plugin for. A list of these plugins and more information
+about using them can be found <a href="/docs/using.html#dns-plugins">here</a>.
+To install one of these plugins, run the installation command above but replace
+<tt>{{package}}</tt> with <tt>{{dns_package_prefix}}-PLUGIN</tt>, where <tt>PLUGIN</tt>
+is the name of the plugin you want to install.
```
bmw: I'm worried that people will think the name of the plugin is "certbot-dns-whatever" because that's what's listed at https://certbot.eff.org/docs/using.html#dns-plugins. They'll probably realize they shouldn't try and install something like "python-certbot-dns-certbot-dns-cloudflare", but can we better clarify what we mean here?

_scripts/instruction-widget/install.js
```diff
@@ -23,6 +23,7 @@ module.exports = function(context) {
 
     context.cron_included = false;
     context.dns_plugins = false;
+    context.dns_package_prefix = "";
```
bmw: It's probably not worth worrying about right now, but it'd be nice if the build failed if we tried to generate DNS installation instructions for a system but hadn't set this. I tried just leaving it unset and building the docs hoping it'd blow up trying to use the unset variable but it didn't work.

Up to you if you want to try and figure this out (maybe by pinging someone like Vivian from E&D about it) or just make an issue for us to look into later.

sydneyli: I'll spin an issue out for this.

_scripts/instruction-widget/templates/install/dnsplugins.html
sydneyli: I added an example using the `rfc2136` plugin!

PR #406, opened 2019-04-16
Clarify the certbot-auto installation instructions

joohoi:
Specify a full path under `/usr/local/bin` for user to copy `certbot-auto` script to, and set proper permissions.


_scripts/instruction-widget/templates/install/commonauto.html
```diff
@@ -2,7 +2,9 @@
 Since it doesn't seem like your operating system has a packaged version of Certbot, you should use our certbot-auto script to get a copy:
 </p>
 <pre>wget https://dl.eff.org/certbot-auto
```
bmw: What are the default file permissions for the file created by `wget`? I think we should verify this file is **always** created with permissions we're OK with (e.g. not world writeable) or change the `chmod` command to `sudo chmod 755 ...`.

_scripts/instruction-widget/templates/install/commonauto.html
```diff
@@ -2,7 +2,9 @@
...
-chmod a+x certbot-auto</pre>
+sudo mv certbot-auto /usr/local/bin/certbot-auto
+sudo chown root:root /usr/local/bin/certbot-auto
```
bmw: I think we should use `chown 0:0` here. `chown 0:0` looks a little odd and I've personally never seen it used before, however, there's not a `root` group on the BSD systems I looked at. Instead, there's a `wheel` group with GID 0.

joohoi: I'm afraid that this might confuse people, as I haven't seen that used either. I'd propose to instead `chown root /usr/local/bin/certbot-auto`, and to `chmod 0755 ...` (as a solution to other comment as well). That ensures that the group wouldn't have the write permissions anyway.

_scripts/instruction-widget/templates/install/commonauto.html
joohoi: I'm up to changing the chmod to octal mode. More context in the [comment below](https://github.com/certbot/website/pull/406#discussion_r276008145)

bmw: Works for me!

_scripts/instruction-widget/templates/install/commonauto.html
bmw: Works for me!

PR #408, opened 2019-04-29
Move freebsd from py27 to py36

apearson:
Freebsd has packages for both python3.6 and py36-certbot(-dns).

Would anyone be opposed to moving the instructions to python 3.6 instead of 2.7?


mfb:
👍 I think now would be a good time to do this, as on April 10th, FreeBSD's default version of python was switched to 3.6: https://svnweb.freebsd.org/ports?view=revision&revision=498529

PR #432, opened 2019-07-01
Completed Infomaniak Certbot informations

ajaegers:


ohemorange:
Hey @ajaegers! Looking at the link you provided, I see it says "a free SSL certficate can be enabled from the console with one simple click" -- requiring a click to enable is what we define as "partial" https support. Have you considered enabling SSL certificates by default?

ajaegers:
Hi @ohemorange thanks for your reply.
This article https://news.infomaniak.com/en/free-lets-encrypt-ssl-certificates-with-infomaniak/ was made long time ago. Since that, we enabled https by default with issuing a Certbot certificate for each websites.
We'll update article to follow reality.

ohemorange:
Updating the article with the current information is great, thank you for making this change! Let me know when the article is updated and we can get this PR in.

ajaegers:
Hello @ohemorange we updated article mentioning that Let's Encrypt is installed by default.

ajaegers:
🎉 👍

PR #449, opened 2019-07-19
Added thexyz

ptoone:


maximillianh:
Thanks for offering Let's Encrypt certificates and for opening this PR! It looks like some configuration is required to enable SSL for thexyz customers, and we list providers as offering 'full' support if they enable SSL by default. 

Can you update the PR to `"category": "full"`? We can move it to the 'full support' section if thexyz's default changes. Thanks!

ptoone:
Okay thanks. I think I have edited that correctly?

ohemorange:
Closing in favor of #451.

PR #456, opened 2019-07-29
Don't fail on broken hosting provider links

### LINKED ISSUE #455: Fix uptime web hosting link
Website builds are failing at https://travis-ci.com/certbot/website/builds/120894065 due to https://github.com/certbot/website/blob/89932f8626b753348c4064084b140e20aad729fa/_data/hosting_providers.json#L1766 404ing.

Quickly looking around their site, I found https://uptimewebhosting.com.au/cpanel-website-hosting-plans/ which if you read carefully says you get can free certs from Let's Encrypt, but I couldn't find anything that says it still happens by default like it did at https://web.archive.org/web/*/https://uptimewebhosting.com.au/knowledgebase/5/Securing-your-website-with-cPanel-web-hosting-plans.html.

@ohemorange, do you have a preferred approach to fix issues like these?

bmw:
This fixes #455 by causing our tests to not check any of the links in the page at https://certbot.eff.org/hosting_providers/.

This is a slightly larger scope than we want, but I'm not aware of a way to only ignore links in the table itself. I personally think this is fine though.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "description": "Gulp build certbot.",
   "scripts": {
-    "test": "htmlproofer ./_site --empty-alt-ignore true --only-4xx true --file-ignore /_site/docs/_modules/,./_site/docs/api/display.html,./_site/docs/contributing.html,./_site/docs/genindex.html,./_site/docs/intro.html,./_site/docs/py-modindex.html,./_site/docs/search.html,./_site/docs/api/log.html,./_site/hosting_providers/index.html --allow-hash-href true"
+    "test": "htmlproofer ./_site --checks-to-ignore LinkCheck --empty-alt-ignore true && htmlproofer ./_site --checks-to-ignore ImageCheck,ScriptCheck --only-4xx true --file-ignore /_site/docs/_modules/,./_site/docs/api/display.html,./_site/docs/contributing.html,./_site/docs/genindex.html,./_site/docs/intro.html,./_site/docs/py-modindex.html,./_site/docs/search.html,./_site/docs/api/log.html,./_site/hosting_providers/index.html --allow-hash-href true"
   },
   "dependencies": {
     "browser-sync": "^2.19.0",
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "description": "Gulp build certbot.",
   "scripts": {
-    "test": "htmlproofer ./_site --empty-alt-ignore true --only-4xx true --file-ignore /_site/docs/_modules/,./_site/docs/api/display.html,./_site/docs/contributing.html,./_site/docs/genindex.html,./_site/docs/intro.html,./_site/docs/py-modindex.html,./_site/docs/search.html,./_site/docs/api/log.html --allow-hash-href true"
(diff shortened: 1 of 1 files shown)


bmw:
Another downside of this approach is it also disables the image and script checks on this file which are described at https://github.com/gjtorikian/html-proofer#whats-tested.

I'm going to briefly timebox myself to see if I can find a nice way to keep these checks on this file.

bmw:
Checks are split. We still have the problem I described in my initial post, but I personally think it's OK.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "description": "Gulp build certbot.",
   "scripts": {
-    "test": "htmlproofer ./_site --empty-alt-ignore true --only-4xx true --file-ignore /_site/docs/_modules/,./_site/docs/api/display.html,./_site/docs/contributing.html,./_site/docs/genindex.html,./_site/docs/intro.html,./_site/docs/py-modindex.html,./_site/docs/search.html,./_site/docs/api/log.html --allow-hash-href true"
+    "test": "htmlproofer ./_site --checks-to-ignore LinkCheck --empty-alt-ignore true --file-ignore ./_site/docs/search.html && htmlproofer ./_site --checks-to-ignore ImageCheck,ScriptCheck --only-4xx true --file-ignore /_site/docs/_modules/,./_site/docs/api/display.html,./_site/docs/contributing.html,./_site/docs/genindex.html,./_site/docs/intro.html,./_site/docs/py-modindex.html,./_site/docs/api/log.html,./_site/hosting_providers/index.html --allow-hash-href true"
```
ohemorange: I'm surprised at this command, because I would have expected it to be:

```
"htmlproofer ./_site --checks-to-ignore LinkCheck --empty-alt-ignore true --file-ignore ./_site/hosting_providers/index.html && htmlproofer ./_site --only-4xx true --file-ignore /_site/docs/_modules/,./_site/docs/api/display.html,./_site/docs/contributing.html,./_site/docs/genindex.html,./_site/docs/intro.html,./_site/docs/py-modindex.html,./_site/docs/api/log.html,./_site/docs/search.html --allow-hash-href true"
```

Could you explain:
- why we're changing how we deal with `search.html`
- what the `--checks-to-ignore` flag does
- if/why we're now checking links on the files that were previously ignored

bmw: > what the --checks-to-ignore flag does

This flag skips the tests specified in the argument to the flag. If you look at the output https://travis-ci.com/certbot/website/builds/121136116#L1270, the first run of `htmlproofer` checks scripts and images and the second checks links.

> if/why we're now checking links on the files that were previously ignored
 
The only change here is to search.html. Explanation of why is below.

> why we're changing how we deal with search.html

The change being made here is we're now checking links on the file. The tests were failing on this file due to a problem with the scripts in the file which can be seen at https://travis-ci.com/certbot/website/builds/121134061#L1278.

Now that we have more granularity in which tests are ignored in which files, we can check links here too but if you prefer, I can also add the file to be ignored in the second command.

bmw: A related change here is that we're now checking images and scripts on all the files that are only ignored in the second `htmlproofer` run.

I can duplicate this list and also ignore the files in the first run, but I'd personally prefer not to. I like tests.

ohemorange: Oh -- does `--checks-to-ignore` ignore the checks for ALL files? I was assuming it means "when you're skipping checks in the files listed in `--files-to-ignore`, these are the checks you should skip".

bmw: > does --checks-to-ignore ignore the checks for ALL files? 

That's my understanding based on https://github.com/gjtorikian/html-proofer/blob/405051645853a6d5df4255aebed957612b137129/bin/htmlproofer#L23 and the output in my previous comment when it prints the checks being run.

I believe if you include a file in `--files-to-ignore`, all checks on that file are skipped.

ohemorange: Gotcha, thanks!

ohemorange:
> This fixes #455 by causing our tests to not check any of the links in the page at https://certbot.eff.org/hosting_providers/.

To say it explicitly: I also think this is fine.

Issue #444, opened 2019-07-09
Add instructions for Debian 10 buster


bmw:
For now, I think we should tell people to just install the packages from the official repos until packages in `buster-backports` are created.

PR #459, opened 2019-08-13
Add debian 10 instructions

bmw:
Fixes https://github.com/certbot/website/issues/444.

### DIFF
--- a/_data/inputs.json
+++ b/_data/inputs.json
@@ -41,10 +41,16 @@
             "version": "9"
         },
         {
+            "name": "Debian 10 (buster)",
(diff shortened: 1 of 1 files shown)


ohemorange:
If we do this, will we have to special-case debian 10 and add an `apt-get remove` like we do for jessie? Or will installing a package of the same name from a different repo Just Work?

bmw:
Assuming you're referring to when we update the instructions to use `buster-backports`, it'll Just Work. Debian sets things up so the packages in that repository are seen as an upgrade the packages this PR tells you to install and the old ones will be removed as part of the installation.

PR #464, opened 2019-08-31
Add Aquatis to host list

AmiT177:


_data/hosting_providers.json
```diff
@@ -2208,5 +2208,15 @@
     "plan": "",
     "reviewed": "2019.7.22",
     "note": ""
-  }
+  },
+  {
```
ohemorange: Order matters; you'll want to add this entry alphabetized in the right section.

_data/hosting_providers.json
```diff
@@ -2208,5 +2208,15 @@
...
+    "name": "Aquatis",
+    "link": "https://aquatis.host/",
+    "category": "full",
+    "tutorial": "",
+    "announcement": "",
+    "plan": "",
+    "reviewed": "2019.8.31",
+    "note": ""
```
ohemorange: Do you provide certificates for all of your hosting products? If not, we may want to add a note saying which products are supported.

_data/hosting_providers.json
```diff
@@ -2208,5 +2208,15 @@
...
     "reviewed": "2019.7.22",
     "note": ""
-  }
+  },
+  {
+    "name": "Aquatis",
+    "link": "https://aquatis.host/",
+    "category": "full",
+    "tutorial": "",
+    "announcement": "",
```
ohemorange: Is there a page on your site that says that you support TLS by default? We should link to that.

_data/hosting_providers.json
AmiT177: Yeah, all of the hosting products have https included.

_data/hosting_providers.json
AmiT177: On the product page it shows up, didn't know if I should link it or not.
https://you.aquatis.host/cart.php?gid=2

ohemorange: hm, you're right, no need. just to check -- there's no steps users need to take or buttons to click to get https support?

AmiT177: Yap, they should get https automatically. (we have AutoSSL with cPanel)

_data/hosting_providers.json
ohemorange: this note was resolved but I'm still seeing it at the end of the list; is there a commit that needs to be pushed?

PR #476, opened 2019-09-29
Add new hosting provider EvergreenTech

NathanFeeser:


bmw:
Thanks for the PR @NathanFeeser.

Can you tell me a bit more about how this works? What does a user of EvergreenTech have to do to set up TLS on their site, especially if they don't request help migrating their site to your service?

NathanFeeser:
Hey, you're very welcome @bmw. My apologies for the delayed response.

EvergreenTech is a fully managed VPS hosting solution. Each client gets their own VPS and dedicated IP and can have an EGT associate establish TLS for them or some of our more technical clients prefer to establish it themselves.

I hope this helps!

bmw:
For a user to to have someone from EvergreenTech set up HTTPS for them, what do they need to do after purchasing a VPS?

I apologize if these questions seem odd or excessive, I'm just trying to figure out where EvergreenTech falls in the categories defined at https://certbot.eff.org/hosting_providers/. I think that is a little trickier than normal because unlike most of the providers listed there, you are a VPS provider rather than a shared hosting provider where the user has limited access to the server itself.

NathanFeeser:
No worries at all, @bmw ! This makes it a bit easier after re-reviewing that guide. I think EvergreenTech would definitely be considered 'Full HTTPS Support'. By default, we provide an SSL cert preinstalled on their VPS as long as they have given us a domain name.

Thank you very much for looking into this. I appreciate what you do!

NathanFeeser:
You're welcome @bmw. My apologies for the delayed response.

I may have misspoken previously. Users aren't required to give us a domain to establish a cert. It is very easy, and also the default, for someone to **not** give us a domain. Most of our clients have a domain already and request we get everything established for them in association with their domain. I hope this makes sense. I'm open to any changes you might suggest.

PR #480, opened 2019-10-01
wetopi provides full LetsEncrypt support with single click install

sitamet:
wetopi provides full LetsEncrypt support with single click install

### DIFF
--- a/_data/hosting_providers.json
+++ b/_data/hosting_providers.json
@@ -1902,11 +1902,11 @@
   {
     "name": "Wetopi",
     "link": "",
-    "category": "partial",
+    "category": "full",
     "tutorial": "https://wetopi.com/install-free-https-certificate/",
     "announcement": "",
     "plan": "",
-    "reviewed": "2019.6.6",
+    "reviewed": "2019.10.1",
     "note": ""
   },
   {


ohemorange:
Hi @sitamet, thanks for coming by! By our definition, "full support" means 0-click install; "partial support" is for 1 or more clicks.

sitamet:
Hi @ohemorange,
ok, don't get confused. All servers launched at wetopi, by default are launched with its letsencrypt cert when running under subdomains. Then, when you move to your own domain. Its 1-click, Is 1-click to say "I want the cert"

maximillianh:
@sitamet thank you for providing your users with free certificates from Let's Encrypt! From what you've described it sounds like additional steps are necessary to install a certificate for a custom domain. EFF lists services where that is the case under 'partial support'. 

We would be happy to move wetopi to 'full support' if my understanding is incorrect or that changes and you provide certificates by default for custom domains.

JamesTheAwesomeDude:
to be clear: Namecheap files **everything short of opt-out** as "partial"?

bmw:
Closing and reopening to trigger CI.

Issue #489, opened 2019-11-02
Debian backports instructions are vague/confusing


alexzorin:
I've seen two users stumble on this, [most recently in this thread](https://community.letsencrypt.org/t/website-link-for-debian-9-backports-links-to-buster-instead-of-stretch/105361?u=_az).

https://certbot.eff.org/lets-encrypt/debianstretch-other links to https://backports.debian.org/Instructions/ , which only shows instructions for the current stable (Buster), not oldstable (Stretch).

If users copy paste the backports instructions as well as the Certbot instructions, it doesn't work.

PR #490, opened 2019-11-04
Don't tell people to use backports

bmw:
This PR solves two issues by no longer telling people to use backports.

1. The packages in Debian Stretch and Devuan ASCII are newer than the versions in backports. Despite this, our current installation command forces the package to be installed from backports which doesn't have fixes for things such as always using POST-as-GET and people have to separately upgrade their packages to fix these problems.
2. Fixes https://github.com/certbot/website/issues/489.

### DIFF
--- a/_scripts/instruction-widget/install.js
+++ b/_scripts/instruction-widget/install.js
@@ -134,14 +134,6 @@ module.exports = function(context) {
     } else if (context.webserver == "nginx") {
       context.package += " " + "python-certbot-nginx";
     }
-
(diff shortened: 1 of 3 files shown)


ohemorange:
if someone's already installed from backports will the new instructions work? code part looks good.

bmw:
Yep! The normal packages in Debian 9 are seen as an upgrade to the packages in backports. The problem currently is that the inclusion of `-t stretch-backports` forces the installation of the packages from there despite there being a newer versions available.

Issue #386, opened 2018-11-30
Dockerfile cannot be built


bmw:
It looks like Ruby fails to build. The relevant output from `docker build` is:
```
Step 13/42 : RUN echo 'gem: --no-document' >> /usr/local/etc/gemrc &&    mkdir /src && cd /src && git clone https://github.com/sstephenson/ruby-build.git &&    cd /src/ruby-build && ./install.sh &&    cd / && rm -rf /src/ruby-build && ruby-build $RUBY_VERSION /usr/local
 ---> Running in 1b68e5430c29
Cloning into 'ruby-build'...
Downloading ruby-2.3.1.tar.bz2...
-> https://cache.ruby-lang.org/pub/ruby/2.3/ruby-2.3.1.tar.bz2
Installing ruby-2.3.1...

BUILD FAILED (Debian GNU/Linux 9 using ruby-build 20181019)

Inspect or clean up the working tree at /tmp/ruby-build.20181130172419.22
Results logged to /tmp/ruby-build.20181130172419.22.log

Last 10 log lines:
make[2]: Leaving directory '/tmp/ruby-build.20181130172419.22/ruby-2.3.1/ext/openssl'
exts.mk:208: recipe for target 'ext/openssl/all' failed
make[1]: *** [ext/openssl/all] Error 2
make[1]: *** Waiting for unfinished jobs....
installing default nkf libraries
linking shared-object nkf.so
make[2]: Leaving directory '/tmp/ruby-build.20181130172419.22/ruby-2.3.1/ext/nkf'
make[1]: Leaving directory '/tmp/ruby-build.20181130172419.22/ruby-2.3.1'
uncommon.mk:203: recipe for target 'build-ext' failed
make: *** [build-ext] Error 2
The command '/bin/sh -c echo 'gem: --no-document' >> /usr/local/etc/gemrc &&    mkdir /src && cd /src && git clone https://github.com/sstephenson/ruby-build.git &&    cd /src/ruby-build && ./install.sh &&    cd / && rm -rf /src/ruby-build && ruby-build $RUBY_VERSION /usr/local' returned a non-zero code: 1
```
@vbrown608, can you take a look at this when you get a chance? I played around with it for a few minutes and thought the fix might be as simple as installing openssl but that didn't work.

mfb:
I am also seeing this.  Also, filed related issue @ #405 (this entire branch of ruby is no longer supported).

mfb:
```
[...]

PR #513, opened 2019-12-10
New docker configuration to develop the website

adferrand:
Hello!

So related to RHEL/CentOS6 32 bits deprecation, and future Windows documentation, I may need to make non obvious changes on the Certbot Website. Sadly, it seems that the current docker configuration to build the website locally is broken: https://github.com/certbot/website/issues/386

I propose in this PR a new docker configuration, that builds and starts the website in development mode through `gulp watch`, using the website sources mounted in the Docker, so the changes can be seen in a browser and committed directly. 

To build the Docker and start it : `cd dev && docker-compose up`. Then, you can see the website from a browser using the URL `http://localhost:4000`. Changes in the source code make the page be refreshed automatically.


bmw:
https://github.com/certbot/website/pull/678 was written by someone on EFF's Engineering and Design team which also fixes our Docker build instructions. Now that that PR is merged, do you think we should still land something like this or is that PR sufficient?

adferrand:
No my PR was serving the same purpose. I close it.

PR #521, opened 2020-01-15
Update hosting_providers.json

veroservers:
Add "Vero Servers"

### DIFF
--- a/_data/hosting_providers.json
+++ b/_data/hosting_providers.json
@@ -2239,7 +2239,7 @@
     "reviewed": "2019.11.18",
     "note": ""
   },
-    {
+  {
     "name": "Vero Servers",
     "link": "https://www.veroservers.com",
     "category": "partial",
--- a/_data/hosting_providers.json
+++ b/_data/hosting_providers.json
@@ -2247,7 +2247,7 @@
     "announcement": "https://twitter.com/veroservers/status/1008697109749649408",
     "plan": "",
     "reviewed": "2020.01.15",
-    "note": ""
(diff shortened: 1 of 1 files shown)


_data/hosting_providers.json
```diff
@@ -2238,6 +2238,16 @@
     "plan": "",
     "reviewed": "2019.11.18",
     "note": ""
+  },
+    {
```
ohemorange:
```suggestion
  {
```

veroservers: Hi,

I guess we should go with "partial" then, because user has to click an option to generate a certificate (on cloud hosting accounts). On managed servers (hostnames) it is setup by default.

ohemorange: Here's my suggestion for how to handle this situation:

- Leave it as "full"
- Add a note to the "note" section saying "Certificates provided by default for managed servers. Cloud hosting accounts offer partial support; see tutorial."
- Add a link to a tutorial or instructions for how to set up a certificate on cloud hosting accounts to the "tutorial" field.

veroservers: I do appreciate your suggestion a lot - will add that tutorial a little later, so for know it could be just left as "partial" and corrected after some time, when tutorial will be in place. Could you accept that?

ohemorange: I'm fine with putting either partial or full as the category, but either way I'd appreciate putting in that note I mentioned specifying which product has which property.

ohemorange: (I'd be happy to accept it as full with a tutorial to come later, as long as there's a note in there.)

PR #525, opened 2020-01-27
Add https://xmission.com to hosting_providers.json

von-copec:
I would like to add my companies hosting to the list.

### DIFF
--- a/_data/hosting_providers.json
+++ b/_data/hosting_providers.json
@@ -2288,5 +2288,15 @@
     "plan": "",
     "reviewed": "2020.01.16",
     "note": "cPanel servers use cPanel's certificate system; all other sites use Let's Encrypt."
+  },
+  {
+    "name": "XMission",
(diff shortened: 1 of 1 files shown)


schoen:
Hi @von-copec,

Thanks for supporting Let's Encrypt in your hosting services and letting us know about them!

It would be helpful to see links to an announcement and/or tutorial that help current or prospective customers learn about your support for Let's Encrypt. The tutorial link could perhaps be https://wiki.xmission.com/Lets_Encrypt, if that's the most relevant resource for most of your subscribers.

I also think that your support probably falls under "partial" rather than "full" in the categorization used by this list, unless there have been subsequent changes that make the Let's Encrypt certificate integration more automatic.

von-copec:
Sorry, first merge request. Does it look like I updated those properly?

schoen:
Thanks, that looks good!

PR #529, opened 2020-02-03
Added ICDSoft to the list of providers

icdsoftware:
I added www.icdsoft.com to the list of providers. We support Let's Encrypt on all of our plans since 2016.


schoen:
Yes, according to other conversations we're reserving "full" for the case where users get certificates with no additional interaction or configuration, so your support should be categorized as "partial" here.

icdsoftware:
Hi, @schoen,

Thank you for your time. I have updated the category to partial and updated the link as per your advice.

PR #536, opened 2020-02-07
Remove notifications from non master branches

bmw:
This is the equivalent of https://github.com/certbot/certbot/pull/7733 but for this repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,7 @@ sudo: required
 notifications:
   email: false
   irc:
+    if: branch = master
     channels:
       # This is set to a secure variable to prevent forks from sending
       # notifications. This value was created by installing


ohemorange:
This doesn't seem exactly equivalent? There we have notifications for PRs failing, here it's only master.

bmw:
I think they are equivalent, but only because of other differences between the repos and the Travis configurations. This stuff is a little confusing and I'm not sure what aspects of Travis you've dug into before so try and bear with me.

On the Certbot repo, we configure tests to only be run on certain branches: https://github.com/certbot/certbot/blob/99b1538d0a70986c7e925e8c091b2f36bded9a4b/.travis.yml#L20-L29.

Combining that list with the [condition for notifications](https://github.com/certbot/certbot/blob/99b1538d0a70986c7e925e8c091b2f36bded9a4b/.travis.yml#L281), IRC notifications will only be sent for failures on the branches whose name matches one of:

    - apache-parser-v2
    - master
    - /^\d+\.\d+\.x$/

On this repo (and the `josepy` repo) we have tests run on all branches and we don't have special `apache-parser-v2` or point release branches, so to get the same behavior where we will only get notified if one of our protected branches fails to build, we just send notifications about the `master` branch.

As for pull requests, Travis doesn't send IRC notifications about pull request builds. They have one sentence documenting this at the bottom of https://docs.travis-ci.com/user/notifications/#configuring-irc-notifications.

The notifications you've probably seen about failures on PR branches were because the branch was named something matching one of the patterns above so in addition to Travis' run for the PR, it also runs tests on the branch as well. You can see an example of this at https://github.com/certbot/certbot/pull/7710 where you can see a [branch build](https://github.com/certbot/certbot/pull/7710/checks?check_run_id=402005982) and a [PR build](https://github.com/certbot/certbot/pull/7710/checks?check_run_id=403875959).

ohemorange:
> The notifications you've probably seen about failures on PR branches were because the branch was named something matching one of the patterns above so in addition to Travis' run for the PR, it also runs tests on the branch as well.

Ah yes, this is the part I was missing, thanks!

PR #538, opened 2020-02-15
adding provider

radeknoone:
Adding the MyDreams innovatisons s.r.o. provider to list.

### DIFF
--- a/_data/hosting_providers.json
+++ b/_data/hosting_providers.json
@@ -1230,6 +1230,16 @@
     "note": ""
   },
   {
+    "name": "MyDreams innovations s.r.o.",
+    "link": "https://www.mydreams.cz/",
+    "category": "full",
+    "tutorial": "",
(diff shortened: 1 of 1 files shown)


schoen:
Hi @radeknoone,

Thanks for this submission!

The page you linked says that users have to take an action in the control panel in order to take advantage of the free certificate; is that correct? If so, we would categorize that as "partial" rather than "full".

radeknoone:
Hi @schoen,

thank you to take the time to check my commit. The information in the link was wrong and now are corrected. We are offering webhosting on the ISPConfig platform with enabled the 'Let's Encrypt' / SSL option by default. Every customer have automatically LE enabled.

Thanks for review in advance!

radeknoone:
Hi all,

may I ask you somebody to merge this commit please? Thank you...

radeknoone:
Hi @bmw!

Would you be so kind and tell me how can I add new provider please?

radeknoone:
@schoen: thanks!

PR #542, opened 2020-02-27
Update hosting_providers

carterdouglasco:
Update of SSL inclusive hosting provider.

### DIFF
--- a/_data/hosting_providers.json
+++ b/_data/hosting_providers.json
@@ -2320,6 +2320,16 @@
     "note": "Shared Web Hosting utilizes Plesk Panel interface to enable Let's Encrypt; Managed Cloud Hosting enabled via Certbot on request."
   },
   {
+    "name": "Marketing One",
+    "link": "https://www.marketingone.com.au",
+    "category": "full",
+    "tutorial": "https://secure.carterdouglas.co/index.php/knowledgebase/12/How-to-Install-Lets-Encrypt-SSL-Using-cPanel.html",
+    "announcement": "",
+    "plan": "",
+    "reviewed": "2020.02.27",
+    "note": "All Shared & Managed Web Hosting offerings include SSL certificates free of charge."
+  },
+  {
     "name": "signaltransmitter.de",
     "link": "https://signaltransmitter.de",
     "category": "partial",


_data/hosting_providers.json
```diff
@@ -2319,6 +2319,16 @@
     "reviewed": "2020.01.29",
     "note": "Shared Web Hosting utilizes Plesk Panel interface to enable Let's Encrypt; Managed Cloud Hosting enabled via Certbot on request."
   },
+  {
+    "name": "Marketing One",
+    "link": "https://www.marketingone.com.au",
+    "category": "full",
```
bmw: Looking at the link you provided, I see users have to manually run AutoSSL. Requiring a click to enable TLS is what we define as "partial" using the categories defined at https://certbot.eff.org/hosting_providers/. Have you considered automatically enabling TLS for your users by default?

carterdouglasco: Hi BMW, 

A script is run upon provisioning of the hosting account which automatically begins to generate the SSL certificate. So yes, we already have TLS enabled by default for all customers.

bmw: That's great!

Do you have any documentation you can link to that says that?

bmw: Also, is the current link on how to do it manually still needed?

carterdouglasco: Apologies for the delay!

We've updated the [article](https://secure.carterdouglas.co/index.php?rp=/knowledgebase/12/How-to-Install-Lets-Encrypt-SSL-Using-cPanel.html) to inform the customers that the SSL certificate are automatically installed for them.

Hope this helps.

Kindly let us know if you require any further information from us.

carterdouglasco: Apologies @bmw , we were not sure what information you required. If you feel as though it may be confusing to customers by including the URL, then we are happy to remove it. However, we stand clear to all customers that SSL is provided free of charge and is automatically installed during the provisioning process.

bmw: That update is great! Thanks @carterdouglasco.

All I was really after was documentation stating that TLS is enabled by default for the purposes of us listing your service on our site.

I personally think the new text should make sense for new users, but if you feel differently or decide you want to change things here in the future, don't hesitate to open a new PR.

I'm going to merge this for now though. Thanks again!

PR #545, opened 2020-03-20
Replace nonunix references with Windows

bmw:
We got an email pointing out to us that our instructions for Windows are currently our instructions for UNIX systems. See https://certbot.eff.org/lets-encrypt/windows-other. This happened because the Windows dropdown used to be "non-UNIX" and the instruction generator code wasn't updated to use the new name.

This PR fixes that.


bmw:
Thanks for the review!

PR #553, opened 2020-04-02
Correct IRC channel

bmw:
This updates the IRC channel used in our privacy policy.

Maybe we also want to update this to reference our public mattermost instance? If so, I  can do that in this PR or we can do it later.

@maximillianh, this change is trivial, but since it's technically a change in our privacy policy, maybe we should have a lawyer take a look? I think you'll have better judgement here than me.


maximillianh:
No need for clearance from Legal on this change.

Issue #492, opened 2019-11-08
Change CentOS/RHEL 8 instructions to use EPEL 8 when packages exist


gopalraha:
With reference to https://github.com/certbot/certbot/issues/7202

bmw:
I made progress on this in #557, but I closed the PR for now until the need for the PowerTools repo is addressed either through documentation or https://github.com/certbot/certbot/issues/7886.

FelixSchwarz:
just fyi: Somehow the acme test suite does not pass in our build environment (https://bugzilla.redhat.com/show_bug.cgi?id=1834530) but everything works fine in a regular virtualenv. I assume this is a due to an "insecure" test certificate and some RHEL hardening.

I'll try to debug this problem in the next days but just wanted to say that your mock removal is really appreciated. certbot 1.4 is on its way for Fedora users at least (CentOS 7 needs extra love because of a Red Hat packaging error: https://bugzilla.redhat.com/show_bug.cgi?id=1834528).

gopalraha:
Hi there, 

Now the Certbot RPM package is available on default CentOS 8 Repository.

Seems like issue has been fixed.

If possible close this ticked and mark it as a resolved

Thanks and regards,
Gopal Raha

bmw:
This ticket hasn't been closed yet because I'm not sure the concerns in my comment above have been addressed yet.

Certbot 1.4.0 should be packaged in EPEL 8 soon, but it also looks like https://fedoraproject.org/wiki/EPEL#Quickstart was updated to tell people to install the `PowerTools` repo addressing the problem through documentation

@FelixSchwarz, do you know if the instructions there to enable EPEL and the `codeready-builder-for-rhel-8-*-rpms` repo will allow people to install Certbot on RHEL 8 or are there other repositories that are required right now?

FelixSchwarz:
The `PowerTools` reference is present in the wiki [since September 2019](https://fedoraproject.org/w/index.php?title=EPEL&type=revision&diff=553585&oldid=552455). When that repo is enabled EPEL's certbot 1.3 installs fine and I'm using it without any problems.

In practice many users do not enable the `PowerTools` repo because it is not needed so they are not aware of that repo at all (at least I was not until I tried to package certbot 1.3 and needed `python3-mock`).

certbot 1.4 will be available in the testing repo soon and should hit stable in ~2 weeks.

[...]

PR #557, opened 2020-04-06
Have CentOS/RHEL 8 users use EPEL 8

bmw:
Fixes https://github.com/certbot/website/issues/492. See https://github.com/certbot/certbot/issues/7202#issuecomment-609385280.

As part of this PR, I generalized our EPEL instructions since the instructions are different for EPEL 7 and EPEL 8 (and our EPEL 7 instructions were out-of-date according to https://fedoraproject.org/wiki/EPEL#Quickstart).

If we wanted to, we could copy the instructions for each EPEL version to our site in this or another PR which may make things easier to understand at the expense of maintainability since we'll have to keep the instructions updated with any new instructions from Fedora.

I also removed the section about installing EPEL when you're not using OS packages rather than also updating the links there since it's dead code since we stopped using EPEL 6 in certbot-auto.

### DIFF
--- a/_scripts/instruction-widget/install.js
+++ b/_scripts/instruction-widget/install.js
@@ -89,8 +89,8 @@ module.exports = function(context) {
   centos_install = function() {
     template = "centos";
 
-    // Certbot only has packages available for RHEL 7 based systems.
-    if (context.version != 7) {
+    // Certbot only has packages available for RHEL 7+ based systems.
+    if (context.version < 7) {
       context.base_command = "/usr/local/bin/certbot-auto";
       // RHEL/CentOS 6 32 bits distros are not supported by certbot-auto.
       if (context.version == 6) {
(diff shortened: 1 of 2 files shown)


FelixSchwarz:
Maybe it would be helpful to mention the PowerTools repo explicitely (for EPEL-8)? In the past I never enabled that package for RHEL 8 machines but we need to do that due to `certbot.tests`/`mock`.

bmw:
Oh thanks for flagging that. I'm closing this PR for now until that's addressed.

I'd like to remove the need to enable the PowerTools repo though. I think you were probably right in https://github.com/certbot/certbot/pull/7755 that Certbot can trivially change to `unittest.mock` in Python 3 as long as all plugins do the same.

If that's correct, would it work for you if we made the `mock` dependency optional in all of our packages in Python 3 and had them automatically fall back to using `unittest.mock` if `mock` isn't available? I think this should remove the need for the dependency while keeping things stable for others who may be relying on our test utilities using the 3rd party `mock` library.

FelixSchwarz:
> If that's correct, would it work for you if we made the mock dependency optional in all of our packages in Python 3 and had them automatically fall back to using unittest.mock if mock isn't available?

Yes, that should be fine.

Issue #443, opened 2019-07-08
How to test built versions of the site?


bmw:
The instructions at https://github.com/certbot/website#travis-builds no longer seem to be accurate. At the very least, I'm getting problems because the URL we send users to for built instructions no longer ends in `.html`.

It might be due to the same problem, but I also can't get the wildcard instructions to load if I manually add `.html` to the URL.

What is the recommended way to see built versions of the site nowadays? @ohemorange, do you know or at least know who we can ask?

We should update the README with the new instructions.

ohemorange:
I actually don't know, I wasn't able to get any version of building the site working. @esoterik do you know?

esoterik:
I haven't actually used the builds repo to view the site -- I was running everything locally without docker for development, set up with the [readme instructions](https://github.com/certbot/website#without-docker)

bmw:
I seem to have found a way to get the site running locally on port 12345 but it's not pretty. Check out the commit from the website-builds repo that you want to play with in your current working directory and run:
```
docker run -p 12345:4000 --rm -v /path/to/unbuilt/website/repo/nginx.conf:/etc/nginx/conf.d/default.conf:ro -v $(pwd):/usr/share/nginx/html:ro -it nginx
```
Before the redesign it seems we didn't need this [try_files](https://github.com/certbot/website/blob/99b24ca05bdc7b2dccff350dbfef70e4b59f0002/nginx.conf#L9) value, at least for the functionality I tested, but now we do (which is totally fine!).

I suppose `gulp` handles this automatically?

PR #558, opened 2020-04-08
Add basic snap installation instructions

bmw:
~~This PR adds instructions for how to install the Certbot snap to our instruction generator. This PR should not be merged until issues like https://github.com/certbot/certbot/issues/7670 are resolved as currently there is no Certbot snap in the beta channel.~~

~~I also think we should merge https://github.com/certbot/website/pull/556 before this PR. While `snapd` does set up a systemd timer to handle renewals, I'm not sure how it handles things like user changes to the file between upgrades. We could look into this and decide that telling people to modify the systemd files like we have been is OK, but I'd personally prefer to land #556.~~

People may want to manually test our snap and these instructions, but I think this issue is unblocked.

### DIFF
--- a/_scripts/instruction-widget/install.js
+++ b/_scripts/instruction-widget/install.js
@@ -258,8 +258,8 @@ module.exports = function(context) {
     template = "snap";
     context.base_command = "certbot";
     context.cron_included = true;
-    context.install_command = "sudo snap install --beta --classic"
-    context.package = "certbot"
+    context.install_command = "sudo snap install --beta --classic";
+    context.package = "certbot";
   }
 
   auto_install = function() {
--- a/_scripts/instruction-widget/install.js
+++ b/_scripts/instruction-widget/install.js
@@ -78,6 +78,7 @@ module.exports = function(context) {
     partials.installcertbot = require(TEMPLATE_PATH + "installcertbot.html");
     partials.dnsplugins = require(TEMPLATE_PATH + "dnsplugins.html");
     partials.dnspluginssetup = require(TEMPLATE_PATH + "dnspluginssetup.html");
+    partials.ssh = require(TEMPLATE_PATH + "ssh.html")
 
     // Load and render the selected template.
     template = require(TEMPLATE_PATH + template + '.html');
--- a/_scripts/instruction-widget/install.js
+++ b/_scripts/instruction-widget/install.js
@@ -32,6 +32,9 @@ module.exports = function(context) {
         context.distro == "sharedhost") {
         return '';
     }
+    else if (context.distro == "snap") {
+      snap_install();
+    }
     else if (context.distro == "debian" && context.version > 8) {
       debian_install();
     }
@@ -249,6 +252,12 @@ module.exports = function(context) {
     }
     context.dns_plugins = true;
     context.dns_package_prefix = "python-certbot-dns";
+  }
+
+  snap_install = function() {
+    template = "snap";
+    context.base_command = "certbot";
+    context.install_command = "sudo snap install certbot --beta --classic"
   }
 
   auto_install = function() {
--- a/_scripts/instruction-widget/install.js
+++ b/_scripts/instruction-widget/install.js
@@ -257,7 +257,8 @@ module.exports = function(context) {
   snap_install = function() {
     template = "snap";
     context.base_command = "certbot";
-    context.install_command = "sudo snap install certbot --beta --classic"
+    context.install_command = "sudo snap install --beta --classic"
+    context.package = "certbot"
   }
 
   auto_install = function() {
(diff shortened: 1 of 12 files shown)


ohemorange:
I plan to test the instructions, but just to check the website generation, do you by any chance have a generated version you could screenshot? I'm pretty sure that last I checked, my machine is not set up to compile the website.

bmw:
Awesome thanks!

A screenshot of our certonly instructions is
Screen Shot 2020-04-21 at 1 58 50 PM

The Apache and Nginx instructions also correctly have you use the Apache and Nginx plugins rather than also using standalone/webroot.

(By the way, I think we should make it easier to test built versions of the site again, but the way I got this screenshot isn't too bad if you want to manually play around with this PR. I used the Travis builds described [here](https://github.com/certbot/website#travis-builds), but used the nginx docker command from my comment [here](https://github.com/certbot/website/issues/443#issuecomment-509436247) instead of `python -m SimpleHTTPServer`.)

_scripts/instruction-widget/templates/install/snap.html
```diff
@@ -0,0 +1,36 @@
...
+    <p>
+    You'll need to install snapd and make sure you follow any instructions to enable classic snap support.<br>
+    Follow these instructions on <a href='https://snapcraft.io/docs/installing-snapd/'>snapcraft's site to install snapd</a>.
+    </p>
+    <p class="centered">
+    <a class="link-button" href='https://snapcraft.io/docs/installing-snapd/'>install snapd</a>
+    </p>
+</li>
+<li>
+    Remove any Certbot OS packages
```
ohemorange, line 26: Can we give examples of how to uninstall certbot for some common cases of this?

_scripts/instruction-widget/install.js
```diff
@@ -251,6 +254,14 @@ module.exports = function(context) {
     context.dns_package_prefix = "python-certbot-dns";
   }
 
+  snap_install = function() {
+    template = "snap";
+    context.base_command = "certbot";
+    context.cron_included = true;
+    context.install_command = "sudo snap install --beta --classic";
+    context.package = "certbot";
```
ohemorange, line 262: just to check, we don't have separate packages for apache or nginx, they're all in there?

_scripts/instruction-widget/install.js
```diff
@@ -251,6 +254,14 @@ module.exports = function(context) {
...
+  }
```
ohemorange, line 263: and I assume we don't have dns plugins ready?

ohemorange: follow-up, what does it look like on the wildcard tab? is there a note about how snaps won't let you get a wildcard cert?

ohemorange: found the answer to the former in the release announcement draft.

bmw: The wildcard tab looks the same as other systems that don't have DNS plugins available. Here's a screenshot:
Screen Shot 2020-04-21 at 3 38 45 PM

_scripts/instruction-widget/install.js
bmw: Yep!

_scripts/instruction-widget/templates/install/snap.html
bmw: Good idea!

Do you have any strong opinions on styling/structure here? Ideas I had are:

* Just list a couple examples inline in the paragraph
* Do an unordered list like we do for `/etc/crontab/`, `/etc/cron.*/*`, etc. at https://certbot.eff.org/lets-encrypt/ubuntubionic-other
* Maybe try to make multiple blocks starting with the command prompt`$`? I putting them all in one block is a bad idea.

ohemorange: either of the first two sounds good to me! agree not to do the last, unless we really want to try to carefully craft a single one that works across machines, but that seems not worth it. I think people tend to know which installation system they have available.

bmw: I played with it a bit and like the last one the best. It looks like 
Screen Shot 2020-04-21 at 4 24 11 PM

_scripts/instruction-widget/templates/install/snap.html
```diff
@@ -29,7 +29,9 @@ <h3>Snap Support</h3>
     <tt>apt</tt>, <tt>dnf</tt>, or <tt>yum</tt>, you should remove them before
     installing the Certbot snap to ensure that when you run the command
     <tt>certbot</tt> the snap is used rather than the installation from your OS
-    package manager.
+    package manager. The exact command to do this depends on your OS, but
+    common examples are <tt>sudo apt-get remove certbot</tt> or <tt>sudo yum
```
ohemorange, line 33: let's also add `dnf` since we mention it right above?

bmw: Sure. I was a little torn because while RHEL/Fedora is trying to push people towards `dnf` so new users may be familiar with `dnf` and not `yum`, I think `sudo yum remove certbot` will work everywhere that `sudo dnf remove certbot` does.

Being consistent makes sense though.

Issue #501, opened 2019-11-21
Add dropdowns for newer Ubuntu versions


bmw:
Our PPA doesn't currently have packages for Ubuntu 19.10. For now I think we should add new drop downs for "Ubuntu 19.04" and "Ubuntu 19.10+". The former should use the PPA and the latter should just use the packages in Ubuntu's official repos for now.

PR #562, opened 2020-04-16
Fix newer Ubuntu instructions

bmw:
Fixes https://github.com/certbot/website/issues/501.

I think this issue has increased in priority because Ubuntu 20.04 will be out soon and I believe it will not contain Python 2 meaning that our current instructions for the platform (found under the "Ubuntu (other)" dropdown) won't work because certbot-auto will assume that Python 2 is available. Let's tell people to just use the OS packages instead and avoid confusion/bug reports about this.

I kept the `id` as `ubuntuother` to prevent any links to those instructions from breaking.

### DIFF
--- a/_data/inputs.json
+++ b/_data/inputs.json
@@ -64,7 +64,7 @@
         },
         {
             "name": "Ubuntu 19.10+",
-            "id": "ubuntueoan",
+            "id": "ubuntuother",
             "distro": "ubuntu",
             "version": "19.10"
         },
--- a/_data/inputs.json
+++ b/_data/inputs.json
@@ -63,6 +63,12 @@
             "version": "0"
         },
         {
+            "name": "Ubuntu 19.10+",
+            "id": "ubuntueoan",
+            "distro": "ubuntu",
+            "version": "19.10"
+        },
+        {
             "name": "Ubuntu 18.04 LTS (bionic)",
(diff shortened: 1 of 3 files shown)


ohemorange:
So we're changing the old-versions-of-ubuntu option from cb-auto to the PPA. Can older versions of ubuntu actually use PPAs? And I take it that what you mean by the python 2 thing above is that PPAs rely on Python 2?

bmw:
This PR changes the "other" Ubuntu instructions from using certbot-auto to using packages in Ubuntu's official repos. Our PPA only has packages for Ubuntu 16.04, 18.04, and 19.04 (and Ubuntu 19.04 has been EOL'd) and no updates to include packages for other versions of Ubuntu are planned by our PPA maintainers anytime soon.

All non-EOL'd versions of Ubuntu other than 16.04 and 18.04 can use these instructions. I'm personally not particularly worried in trying to continue to support instructions for EOL'd versions, but they should work for all versions starting sometime between Ubuntu 16.04 and Ubuntu 18.04. I'm not immediately sure when the nginx and DNS plugins were added.

In my initial post I'm talking about how certbot-auto tries to install and use Python 2 on Ubuntu which won't work on Ubuntu 20.04 if the `python2` package has been removed.

PR #563, opened 2020-04-20
Update hosting_providers.json

azarainey:
adding porkbun registrar


m0namon:
Hi @azarainey ! Thank you for the addition.
A couple of clarification questions! Since this is a list of hosting providers, we would want to focus on your hosting services. In the tutorial link you provided, I see:

 > Better yet, if you're using Porkbun's hosting package or site builder, you don't have to do anything at all to add SSL security: the sites will install automatically.

Is there an announcement post detailing this? That seems to actually qualify for the "full" support category.

PR #567, opened 2020-04-27
Fix Namecheap hosting tutorial

JamesTheAwesomeDude:
I hope this proves useful to others.

I have attached the "long-winded" explanation of my proposed changes in the commit message; the "tldr" of which is:

* _None_ of the previous links were even _applicable_ to Let's Encrypt.
* Most of them were _only_ applicable to "Namecheap SSL", a branded, paid product from their partner, Codomo.
* I have replaced _all_ these links with both applicable and informative URLs.

### DIFF
--- a/_data/hosting_providers.json
+++ b/_data/hosting_providers.json
@@ -2212,12 +2212,12 @@
   {
     "name": "Namecheap.com",
     "link": "https://www.namecheap.com/hosting/",
-    "category": "partial",
+    "category": "no",
     "tutorial": "https://www.namecheap.com/support/knowledgebase/article.aspx/9418/33/installing-an-ssl-certificate-on-your-server-using-cpanel/",
     "announcement": "",
     "plan": "https://community.letsencrypt.org/t/namecheap-response-to-lets-encrypt/51411",
     "reviewed": "2020.4.27",
-    "note": "automatic renewal via ACME not offically supported"
+    "note": ""
   },
   {
     "name": "Netsons.com",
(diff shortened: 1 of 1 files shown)


JamesTheAwesomeDude:
In fact, I'd be amenable to going so far as setting `"category": "no"` and cancelling the `note` I have "penciled in" there.

I mentioned it in the commit message, but it bears repeating:

Namecheap's "support" for Let's Encrypt is **materially identical** to that of GoDaddy -- who is currently filed as "No". They both provide merely cPanel's access to upload-your-own key+CRT, with some integrated utilities for generating CSRs, etc.

Another strong candidate for the "plan" link might be [this alleged email from Namecheap staff on the topic](https://gist.github.com/stopspazzing/c871f935771f14933c63).

> ...we truly believe purchased SSL is a better choice in most cases...  
>  -`sslsupport@namecheap.com`

🙄

-----

(Note that I am currently, personally _testing_ to see whether or not certbot-manual-in-crontab is capable of updating the `.crt` file "in-place"; if it works, that would indeed change this from an ambiguous-No to a solid-Partial, IMO. But, as it stands, they don't provide _ACME_ support _per se_.)

JamesTheAwesomeDude:
Alas, there's a typo in my last commit. `s/Namecheap cert/Let's Encrypt cert/`

JamesTheAwesomeDude:
> The "tutorial" and "plan" links aren't relevant for hosting services that don't offer partial or full HTTPS support, so you can remove these two links altogether.

I mean, the "tutorial" link shows you how to install third-party certs (such as Let's Encrypt) -- Namecheap **supports HTTPS**, including with custom certs -- just not *ACME*.

m0namon:
Yep! We recognize that-- I should have been more specific, "partial" and "full" as defined by our criteria around automated renewal.

Either way, only one (or none) of "tutorial," "announcement," and "plan" can be filled out -- it determines what shows up in the "Relevant Link" section of the table: https://certbot.eff.org/hosting_providers/

"Tutorial" seems most relevant-- so I'd accept this if both links were removed, or just plan.

Thanks again for the fix.

PR #569, opened 2020-04-28
Make sure universe is enabled in non-PPA Ubuntu instructions

bmw:
As reported at https://github.com/certbot/certbot/issues/7953, our non-PPA Ubuntu instructions don't work if Ubuntu's universe repository isn't enabled. (If you're unfamiliar with the different components of Ubuntu's repos, you can find a short summary at the top of https://help.ubuntu.com/community/Repositories/Ubuntu).

Universe seems to be enabled by default in most places, but this change makes our instructions work even if it's not enabled.

Our non-PPA Ubuntu instructions are unchanged and the new instructions for 19.10+ look like 
Screen Shot 2020-04-28 at 11 26 54 AM

### DIFF
--- a/_scripts/instruction-widget/templates/install/ubuntu.html
+++ b/_scripts/instruction-widget/templates/install/ubuntu.html
@@ -22,6 +22,7 @@
       <li>sudo apt-get install software-properties-common</li>
       <li>sudo add-apt-repository universe</li>
       <li>sudo apt-get update</li></ol></pre>
+    </p>
 {{/ppa}}
 </li>
 
--- a/_scripts/instruction-widget/templates/install/ubuntu.html
+++ b/_scripts/instruction-widget/templates/install/ubuntu.html
@@ -1,7 +1,7 @@
 {{> header}}
 
-<li>
 {{#ppa}}
+<li>
     Add Certbot PPA
     <p>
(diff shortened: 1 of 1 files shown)


_scripts/instruction-widget/templates/install/ubuntu.html
```diff
@@ -12,8 +12,19 @@
...
+    <p>
+    You'll need to make sure Ubuntu universe is in your list of repositories.
+    To do so, run the following commands on the command line on the machine:
+    <pre class="no-before"><ol><li>sudo apt-get update</li>
+      <li>sudo apt-get install software-properties-common</li>
+      <li>sudo add-apt-repository universe</li>
+      <li>sudo apt-get update</li></ol></pre>
+    </p>
+{{/ppa}}
+</li>
```
ohemorange, line 27: Exxxtreeeemely minor style consistency nit: the `li`s are usually within the `{{}}`, even when they're mutually exclusive. It's because the concept is a flexible list that includes or doesn't include items, vs being conceptually a branching tree with different relevant options at the different steps.

bmw: Makes sense! I made that change.

PR #577, opened 2020-05-13
Replace Windows instructions with an adaptation of Adrien's documentation

### LINKED ISSUE #517: Add Windows beta instructions
As suggested by Adrien, I think we should add instructions about the Windows beta to the instruction generator. This should be pretty easy as we can add a "Windows" dropdown option with a prominent warning at the top of the generated instructions saying that Windows support is still in its beta phase.

As part of this, I'm not sure what we should do with our "Non-UNIX" dropdown option. Maybe we should just delete it?

schoen:
This will make the instruction generator give useful instructions on how to use the current version of Certbot for Windows.

Fixes #517.

### DIFF
--- a/_scripts/instruction-widget/templates/getting-started/windows.html
+++ b/_scripts/instruction-widget/templates/getting-started/windows.html
@@ -1,7 +1,112 @@
+<h1>Windows installation procedure</h1>
+
 <p>
-Certbot is currently only available for UNIX-like operating systems. Although
-EFF's Certbot might not work for your use case, there are many <a
-href="https://community.letsencrypt.org/t/list-of-client-implementations/2103">
+Certbot is now officially available for Windows. If you find that Certbot is not
+the most suitable Let's Encrypt client application for your use case, there are
+many <a href="https://community.letsencrypt.org/t/list-of-client-implementations/2103">
 other clients</a> written by other organizations and developers that you may be
 able to use to obtain a certificate from Let's Encrypt.
 </p>
+
+<h2>Important notes</h2>
+
+<p>
+This procedure follows the current Certbot implementation for Windows, in particular the fact that it installs as a system component, and requires administrative privileges. These instructions will be updated when a future version of Certbot <a href="https://github.com/certbot/certbot/issues/7883">switches to a different installation method</a>.
+No installers for HTTP servers are supported for now (Certbot for Windows can currently obtain your certificate from
+Let's Encrypt, but not install it into your web server application).
+
+<h2>Specific Windows system requirements and user knowledge requirements</h2>
+
+<ul>
+<li>The user needs to be familiar with the command-line interface (CLI), because Certbot is a pure CLI program.</li>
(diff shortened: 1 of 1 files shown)


bmw:
Thanks for the review!

@adferrand, I know you took a high level look at this PR at one point and if I remember correctly, you had some changes you wanted to suggest. I don't think you need to do a full review now, but was there anything here you think needs to be fixed before this goes live?

_scripts/instruction-widget/templates/getting-started/windows.html
```diff
@@ -1,7 +1,112 @@
...
+<li>The user must use an account with administrative privileges to install and run Certbot.</li>
+<li>PowerShell and <code>CMD.EXE</code> are supported; both need to be started with elevated privileges before invoking Certbot.</li>
+<li>Path <code>C:\Certbot</code> must be writable by the current user.</li>
+</ul>
+
+<h2>Specific Windows limitations and configuration</h2>
+
+<ul>
+<li>All usual operations to create and manage an account, manage existing certificates, or select the ACME server, are supported.</li>
+<li>Only standalone, manual and webroot authenticator plugins are supported. DNS plugins will be available soon. This means that Certbot for Windows is currently unable to automatically renew wildcard certificates, since these require a DNS plugin in order to be renewed without user intervention.</li>
```
adferrand, line 31: One remark here: even if DNS plugins are not available, wildcard can still be issued if the user implements the proper hooks using the `manual` plugin to communicate with the DNS provider API, and create/clean TXT records on the DNS zone as required.

_scripts/instruction-widget/templates/getting-started/windows.html
```diff
@@ -1,7 +1,112 @@
...
+  If you want to check that you have the top-of-the-line installation, you can head to
+  <a href='https://www.ssllabs.com/ssltest/'>https://www.ssllabs.com/ssltest/</a>.
+  </p>
+  <p class="centered">
+  <a class="link-button" href='https://www.ssllabs.com/ssltest/'>check your site's <img src="/images/Lock.svg"> https:// at SSL Labs</a>.
+  </p>
+
+<h2>Note for Windows Apache or nginx users</h2>
+
+<p>As described in section 5 above, Certbot for Windows currently cannot install the certificate in Apache or nginx for you. As of the most recent release, you will have to edit your web server application’s configuration to install the certificate yourself after Certbot has obtained it. If this limitation is acceptable to you, please start from the beginning of this document to learn more about installing and using Certbot on Windows.</p>
```
adferrand, line 112: nginx => Nginx

maximillianh: Corrected, thank you.

_scripts/instruction-widget/templates/getting-started/windows.html
maximillianh: While we do have a workaround to allow issuance of wildcard certificates I think it's still accurate to say that wildcard certificates are not automatically renewed.

PR #581, opened 2020-05-18
Have CentOS/RHEL 8 users use EPEL 8 instead of certbot-auto

### LINKED ISSUE #492: Change CentOS/RHEL 8 instructions to use EPEL 8 when packages exist

bmw:
Fixes #492.

As part of this PR, I generalized our EPEL instructions since the instructions are different for EPEL 7 and EPEL 8 and our EPEL 7 instructions were out-of-date according to https://fedoraproject.org/wiki/EPEL#Quickstart.

I also removed the section about installing EPEL when you're not using OS packages rather than also updating the links there since it's dead code since we stopped using EPEL 6 in certbot-auto.

### DIFF
--- a/_scripts/instruction-widget/templates/install/centos.html
+++ b/_scripts/instruction-widget/templates/install/centos.html
@@ -14,7 +14,9 @@
     Enable EPEL repo
     <p>
     You'll need to enable the EPEL (Extra Packages for Enterprise Linux)
-    repository.<br>
+    repository and make sure you follow all instructions for your system,
+    including enabling any other recommended repositories that may be
+    required.<br>
     Follow these instructions at
     <a href='https://fedoraproject.org/wiki/EPEL#Quickstart'>
       the Fedora wiki to enable EPEL</a>.
--- a/_scripts/instruction-widget/templates/install/centos.html
+++ b/_scripts/instruction-widget/templates/install/centos.html
@@ -16,22 +16,12 @@
     You'll need to enable the EPEL (Extra Packages for Enterprise Linux)
     repository.<br>
     Follow these instructions at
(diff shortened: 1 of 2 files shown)


_scripts/instruction-widget/templates/install/centos.html
```diff
@@ -16,46 +16,20 @@ <h3>Supported architectures:</h3>
...
       the Fedora wiki to enable EPEL</a>.
     </p>
     <p class="centered">
-    <a class="link-button" href='https://fedoraproject.org/wiki/EPEL#How_can_I_use_these_extra_packages.3F'>
+    <a class="link-button" href='https://fedoraproject.org/wiki/EPEL#Quickstart'>
       enable EPEL</a>
     </p>
 </li>
-<li>
-    Enable the optional channel
```
ohemorange, line 28: Is this no longer needed as well? It sounds like it's maybe still needed for 7?

bmw: It's still needed and the other packages listed at https://fedoraproject.org/wiki/EPEL#Quickstart for RHEL 7 may be needed now or in the future as well.

At least for now, I'd personally rather just link to their instructions which should always stay up to date rather than trying to duplicate them here.

ohemorange: Makes sense. Could we add a note to the effect of "make sure to follow all instructions for your system, including installing optional packages and extra if necessary"?

bmw: Good idea. The `PowerTools`/`codeready-builder` repositories are currently needed for the Certbot packages on RHEL 8 so I made the wording a little more generic, but otherwise added the suggestion.

Issue #588, opened 2020-06-07
Broken markup in Windows instructions


alexzorin:
https://certbot.eff.org/lets-encrypt/windows-other

It looks like something is going wrong with the `<ol>`  under *Installation instructions*.

![markup in text](https://i.imgur.com/WNdLCFh.png)

and

![markup in text and section 5 does not exist](https://i.imgur.com/g3bm1Ux.png)

schoen:
I believe this can be fixed with #590 now—at least the `</ol>` renders properly there now!

alexzorin:
Thanks @schoen!

schoen:
Now even better due to https://github.com/certbot/website/pull/601!

PR #590, opened 2020-06-09
Try balancing some tags

schoen:
Maybe this will fix the rendering issue #588?


schoen:
It looks to me like it works now, oddly enough, and hence

Fixes #588.

PR #595, opened 2020-06-11
Adding snapd text to certain OS instructions

### LINKED ISSUE #559: Add banner pointing people to snapd?
Currently we're planning on suggesting the majority of our users use the Certbot snap in the future because it provides an easy way for most users to get the latest version of Certbot with things like automatic renewal preconfigured.

I added a snap dropdown to the instruction generator at https://github.com/certbot/website/pull/558, but I think it may be worth having the instructions for systems that can use the snap suggest people try it too. Installing Certbot in this way will first go through a beta period which this banner should probably make clear, however, I think it would encourage many more people to try it.

I think it probably makes sense to start the Certbot snap beta without this change and then consider it if not many people are using it.

lschatzkin:
closes #559 
I've added this text to templates for the OSs requested. I tried creating a template partial and adding it that way, the *better* way, but couldn't get it to work. I ended up added it to each relevant template, sorry for the lack of DRY code. 

Also, two OSs aren't included that should be, Debian(other) and Ubuntu(other). Their script generates from the commonauto.html template. I could include the text there, but then it would also show on some OSs that it isn't relevant to.

Here's how the text is appearing, except for Debian 8, where it appears above instruction 2.
Screenshot from 2020-06-11 11-51-41

### DIFF
--- a/_scripts/instruction-widget/templates/install/centos.html
+++ b/_scripts/instruction-widget/templates/install/centos.html
@@ -1,4 +1,11 @@
+{{#packaged}}
+<aside class="note">
+  <div class="note-header">
+    <h3>Join our beta test!</h3>
+  </div>
 <p>If you would like to help test a new way to install Certbot that automatically renews your certificates and stays up-to-date, please choose "snapd" in the system dropdown above.</p>
+</aside>
+{{/packaged}}
 
 {{#deprecated_32bits}}
 <aside class="note">
--- a/_scripts/instruction-widget/templates/install/centos.html
+++ b/_scripts/instruction-widget/templates/install/centos.html
@@ -1,3 +1,5 @@
+<p>If you would like to help test a new way to install Certbot that automatically renews your certificates and stays up-to-date, please choose "snapd" in the system dropdown above.</p>
+
 {{#deprecated_32bits}}
 <aside class="note">
   <div class="note-header">
--- a/_scripts/instruction-widget/templates/install/commonauto.html
+++ b/_scripts/instruction-widget/templates/install/commonauto.html
@@ -1,4 +1,6 @@
 {{#jessie}}
+<p>If you would like to help test a new way to install Certbot that automatically renews your certificates and stays up-to-date, please choose "snapd" in the system dropdown above.</p>
+
 <li>
     Remove packaged Certbot installation
     <p>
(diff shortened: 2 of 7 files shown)


_scripts/instruction-widget/templates/install/commonauto.html
```diff
@@ -1,4 +1,6 @@
 {{#jessie}}
+<p>If you would like to help test a new way to install Certbot that automatically renews your certificates and stays up-to-date, please choose "snapd" in the system dropdown above.</p>
```
bmw, line 2: We don't want to include this for Debian Jessie. They're unable to use the snaps unfortunately.

I think you can just delete the change here.

_scripts/instruction-widget/templates/install/centos.html
```diff
@@ -1,3 +1,5 @@
+<p>If you would like to help test a new way to install Certbot that automatically renews your certificates and stays up-to-date, please choose "snapd" in the system dropdown above.</p>
```
bmw, line 1: We don't want to show this text to CentOS/RHEL 6 users. To fix this, I'd wrap this text with:
```
{{#packaged}}
{{/packaged}}
```

_scripts/instruction-widget/templates/install/debian.html
```diff
@@ -1,3 +1,5 @@
+<p>If you would like to help test a new way to install Certbot that automatically renews your certificates and stays up-to-date, please choose "snapd" in the system dropdown above.</p>
```
bmw, line 1: We don't want these instructions shown for Devuan users. To fix this, I think you can wrap this text with
```
{{^devuan}}
{{/devuan}}
```
but it'd be good to double check this works.

lschatzkin:
> <aside class="note">
>   <div class="note-header">
>     <h3>Join our beta test!</h3>
>   </div>
> <p>If you would like to help test a new way to install Certbot that automatically renews your certificates and stays up-to-date, please choose "snapd" in the system dropdown above.</p>
> </aside>

I think this is a fine idea! I thought it was too subtle but I thought perhaps you wanted it that way.

_scripts/instruction-widget/templates/install/commonauto.html
lschatzkin: Oops, misunderstood, my bad.

_scripts/instruction-widget/templates/install/debian.html
lschatzkin: Looks like it's working to me.

lschatzkin:
> LGTM. Thanks @lschatzkin!

Glad I could help!

PR #607, opened 2020-07-13
Remove version from Fedora dropdown

### LINKED ISSUE #606: "Fedora 26+" should be kept up-to-date with EOLs
Users may get the impression that Certbot is maintained for F26, but nothing is being uploaded or backported. Certbot installs, but fails mysteriously with "method not allowed" or other errors.

I suggest updating it to say "Fedora 31+" and keeping it current with https://fedoraproject.org/wiki/End_of_life.

alexzorin:
Fixes #606, for now.

### DIFF
--- a/_data/inputs.json
+++ b/_data/inputs.json
@@ -93,10 +93,10 @@
             "version": "0"
         },
         {
-            "name": "Fedora 26+",
+            "name": "Fedora 31+",
             "id": "fedora",
             "distro": "fedora",
-            "version": "26"
+            "version": "31"
         },
         {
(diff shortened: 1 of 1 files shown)


ohemorange:
Thanks for catching this, good call! Looks like fedora EOLs every six months, which is kind of annoying to keep up-to-date. What about just removing the version number, like we have for gentoo? Looks like we're not using it anyway.

alexzorin:
Alright, fair enough. It's too bad there's no way to visibly deprecate the old packages.

PR #608, opened 2020-07-15
Add NederHost to hosting_providers.json.

sebastiaanhoogeveen:
NederHost enables HTTPS with Let's Encrypt for all websites by default (and forces redirect to HTTPS for new websites). The announcement is in Dutch; if a translation is required please let me know.


ohemorange:
Hm, looks like this wasn't sent to travis for some reason, going to close and reopen.

PR #609, opened 2020-07-16
Fix RHEL 8 renewal instructions

bmw:
As reported at https://community.letsencrypt.org/t/selinux-should-the-log-files-be-var-log-t-or-cron-log-t/128552/2, our renewal instructions for CentOS/RHEL 8 don't work. We initially fixed this in https://github.com/certbot/website/pull/528 but broke it in https://github.com/certbot/website/pull/581. This PR fixes them again.

### DIFF
--- a/_scripts/instruction-widget/install.js
+++ b/_scripts/instruction-widget/install.js
@@ -88,21 +88,17 @@ module.exports = function(context) {
   centos_install = function() {
     template = "centos";
 
-    // Certbot only has packages available for RHEL 7+ based systems.
+    // Certbot only has packages available for RHEL 7+ based systems and RHEL 6
+    // is the oldest version we support.
     if (context.version < 7) {
       context.base_command = "/usr/local/bin/certbot-auto";
-      // RHEL/CentOS 6 32 bits distros are not supported by certbot-auto.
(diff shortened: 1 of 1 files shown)


ohemorange:
Since we broke this the last time we tried to fix it, would you mind posting what this looks like for 6, 7, 8?

bmw:
Sure. Happy to provide that as part of the PR, but just to make sure you know, all the instructions for running the site locally in our [README](https://github.com/certbot/website/blob/master/README.md) should work now. E&D is also  working on further simplifying the [Travis builds](https://github.com/certbot/website/blob/master/README.md#travis-builds) instructions as well.

CentOS 6:
centos6

RHEL 6:
rhel6

CentOS/RHEL 7
centos:rhel7

CentOS/RHEL 8
centos:rhel8

bmw:
When we were instructing CentOS/RHEL 8 users to use certbot-auto (which was added in https://github.com/certbot/website/pull/458), we removed the EPEL 8 instructions for them since we didn't need any packages from EPEL 8 for certbot-auto to work.

Now Certbot is packaged in EPEL 8 so we're instructing users to enable that repository and install the packages found there.

Thanks for the review!

PR #613, opened 2020-07-24
Document how to install DNS plugin snaps

### LINKED ISSUE #603: Document how to install DNS plugin snaps
~~This work is blocked on https://github.com/certbot/certbot/issues/8041 being resolved.~~

I think these instructions should be added to the "wildcard" tab of the "snapd" dropdown. The instructions at the end of https://github.com/certbot/certbot/blob/d434b92945f04f84b10fee1048cb45a15d3a36ba/tools/snap/README.md#build-the-snaps can be used as a reminder of the commands that need to be run here.

One thing to note though is it appears the DNS plugins will automatically try to connect to Certbot when they are installed so I think we should run `sudo snap set certbot trust-plugin-with-root=ok` before installing the DNS plugin and then the `sudo snap connect` command shouldn't be needed.

I also think:

- [x] We should make sure there is a banner at the top of these instructions saying they are still in the beta phase even once https://github.com/certbot/certbot/issues/8140 is resolved.
- [x] We should make sure that both Certbot and the DNS plugin snap are installed from the same channel. I personally think we should use `--edge` here assuming we tackle this issue before we have DNS plugin snaps in the "beta" channel. If we do this though, I think we should create an issue to track changing the instructions to `--beta` once beta versions exist.

ohemorange:
Fixes #603.

Follow-up issue for after next release in https://github.com/certbot/website/issues/614.

Default:
Screenshot from 2020-07-29 14-28-50

Wildcard:
Screenshot from 2020-07-29 14-28-37


ohemorange:
Nice catch, thanks! Yes, I agree we should fix this. I think I can do that without too much trouble.

ohemorange:
This should be fixed, and screenshots are updated.

bmw:
Welp, it looks like I didn't test things as thoroughly as Alex Zorin. See https://github.com/certbot/certbot/issues/8181.

The TL;DR is I think we should add a step telling users to run `sudo snap connect certbot:plugin <PLUGIN>` after installing the plugin to ensure things work consistently if multiple plugins are installed.

ohemorange:
Wait, what? Why is it doing that? That's very surprising to me. Is something broken?

ohemorange:
Ack, it goes by very quickly, but it says something like "failed to check assertions". I don't think the instructions are the problem, it might be a problem with the version check or something.

ohemorange:
Looks like only the metadata interface is connecting, not the plugin interface.

bmw:
Maybe something is broken, but I just assumed there was undocumented behavior where snapd/snap didn't want to connect two producers to the same consumer over the same content interface. I could see that causing problems in general and I think if we were using the content interface like normal, it'd try to merge the two plugin snap's Python files into the directory the Certbot snap reads from probably would go poorly.

On my system, the output of `sudo snap install` with a 2nd plugin is:
```
$ sudo snap install --edge certbot-dns-digitalocean >out 2>&1; strings out
Fetch and check assertions for snap "certbot-dns-digitalocean" (56)                                                                                                                                                                                                                                                                                                                                                                  /
Fetch and check assertions for snap "certbot-dns-digitalocean" (56)                                                                                                                                                                                                                                                                                                                                                                  -
Fetch and check assertions for snap "certbot-dns-digitalocean" (56)                                                                                                                                                                                                                                                                                                                                                                  \
Fetch and check assertions for snap "certbot-dns-digitalocean" (56)                                                                                                                                                                                                                                                                                                                                                                  |
Fetch and check assertions for snap "certbot-dns-digitalocean" (56)                                                                                                                                                                                                                                                                                                                                                                  /
Fetch and check assertions for snap "certbot-dns-digitalocean" (56)                                                                                                                                                                                                                                                                                                                                                                  -
Fetch and check assertions for snap "certbot-dns-digitalocean" (56)                                                                                                                                                                                                                                                                                                                                                                  \
Mount snap "certbot-dns-digitalocean" (56)                                                                                                                                                                                                                                                                                                                                                                                           |
Mount snap "certbot-dns-digitalocean" (56)                                                                                                                                                                                                                                                                                                                                                                                           /
Mount snap "certbot-dns-digitalocean" (56)                                                                                                                                                                                                                                                                                                                                                                                           -
Mount snap "certbot-dns-digitalocean" (56)                                                                                                                                                                                                                                                                                                                                                                                           \
Mount snap "certbot-dns-digitalocean" (56)                                                                                                                                                                                                                                                                                                                                                                                           |
Mount snap "certbot-dns-digitalocean" (56)                                                                                                                                                                                                                                                                                                                                                                                           /
Mount snap "certbot-dns-digitalocean" (56)                                                                                                                                                                                                                                                                                                                                                                                           -
Mount snap "certbot-dns-digitalocean" (56)                                                                                                                                                                                                                                                                                                                                                                                           \
Setup snap "certbot-dns-digitalocean" (56) security profiles                                                                                                                                                                                                                                                                                                                                                                         |
Setup snap "certbot-dns-digitalocean" (56) security profiles                                                                                                                                                                                                                                                                                                                                                                         /
Setup snap "certbot-dns-digitalocean" (56) security profiles                                                                                                                                                                                                                                                                                                                                                                         -
Setup snap "certbot-dns-digitalocean" (56) security profiles                                                                                                                                                                                                                                                                                                                                                                         \
Setup snap "certbot-dns-digitalocean" (56) security profiles                                                                                                                                                                                                                                                                                                                                                                         |
Setup snap "certbot-dns-digitalocean" (56) security profiles for auto-connections                                                                                                                                                                                                                                                                                                                                                    /
Setup snap "certbot-dns-digitalocean" (56) security profiles for auto-connections                                                                                                                                                                                                                                                                                                                                                    -
[?25h
[Kcertbot-dns-digitalocean (edge) 1.7.0.dev0 from Certbot Project (certbot-eff*) installed
```

ohemorange:
That would be surprising since connecting two producers to one consumer is expected, defined behavior (on the documentation page for the content interface). It looks like the hook isn't getting run, because the option is still set. It's also stopped printing that message, trying to get it back.

ohemorange:
"By default, when multiple directories are shared from a producer snap, or when multiple slots are connected to a single plug, the shared content is merged under the target path of the consuming path’s plug definition."
https://snapcraft.io/docs/content-interface

ohemorange:
I wonder if we're finally hitting the problem that the interface isn't actually set up correctly with proper directory locations

bmw:
Thanks for the link.

Between the output and syslogs, I haven't seen any error output about this. I also noticed:

1. Explicitly connecting the two snaps seems to work just fine.
2. The first plugin install prints `Run hook prepare-plug-plugin of snap "certbot"` while subsequent plugins do not for me.
3. There is no error if you don't run `sudo snap set certbot trust-plugin-with-root=ok` before installing subsequent plugins.

All of that made me suspect it's not even trying to connect on installation, but I certainly don't know that and you said you thought you saw some error output. It's certainly possible there's an underlying problem.

Do you think we should block announcing the DNS plugin beta on this? Do you think this should block announcing the stable Certbot snap too? To investigate this problem, the first thing I think we should try is creating a thread about this on the snapcraft discourse rather than try to debug their software producing no output ourselves. I can do that myself tomorrow if you'd like.

ohemorange:
Yes, agreed about it not trying to connect. I don't know why though. I think we should check on the snapcraft forums, and block announcing the plugins on getting a response. Possibly also the stable certbot, but I feel less strongly about that.

bmw:
I did a little research and created https://forum.snapcraft.io/t/how-do-you-configure-multiple-slots-to-auto-connect-to-a-single-plug/19172.

bmw:
If you've been following that discussion, it appears it is the undocumented behavior that plugs will only auto-connect to the first slot by default and you have to ask the snap store maintainers to change the declarations they add to your snap on in the snap store to change this behavior.

I think we should word the request a little carefully/naively to avoid unintended side effects (such as trying to "greedily" connect to snaps not published by us that define the correct interface), but I think that we should open up a snap store request for this change. Do you agree?

If so, I think we should get that process started. If you have the time and interest to do that, that'd be helpful, but I can do it myself next week too.

If you would like to wait a few days to see if they can quickly fix that to avoid the need for a manual connect step in our instructions here, I think that's fine, but I do not think we should block announcements like the stable Certbot snap much longer. I'd really like to move forward on that to keep working towards killing certbot-auto and Python 2 support.

ohemorange:
Oh, fun. Yes I definitely agree that we should ask them to do this. I think it's reasonable to announce the stable snap, but I think we should wait a little and hold off on further dns plugins announcements. Is there a specific format we need to ask them for this? Or just describe the request in the appropriate category on the forum?

ohemorange:
We could either split up the announcement or wait.

bmw:
> Is there a specific format we need to ask them for this? Or just describe the request in the appropriate category on the forum?

There's no documentation about how to make the request ☹️ There's one example at https://forum.snapcraft.io/t/auto-connect-and-greedy-plug-behaviour-for-for-gtk-common-themes-and-gtk-theme-traditionalhumanized-snaps/19113 but I certainly don't think we should try to write the config for them like they did.

If I write the post next week, in the `store-requests` section I probably would just carefully explain the behavior we want (e.g. "we'd like the plug in the Certbot snap to auto-connect to all matching slots in snaps published by us" and ask if they can help us with that linking to https://forum.snapcraft.io/t/plug-slot-declaration-rules-greedy-plugs/12438 and/or my thread.

> We could either split up the announcement or wait.

I personally think we should make the post and maybe wait up to a week, but if it isn't resolved then we should either add the connect command here or split the announcement.

ohemorange:
I'll go ahead and make that request now.

ohemorange:
Thread is here: https://forum.snapcraft.io/t/request-for-greedy-plug-certbot-plugin-interface/19190

bmw:
Thanks a lot for getting that started!

ohemorange:
Here's the updated instructions. We're in the queue, so that's nice, but we should probably just get this out.
Screenshot from 2020-08-17 17-56-46

bmw:
I think we should go ahead and update the instructions to use the beta channel now. Our DNS plugin snaps have been in the beta channel since our last release and it avoids us ever telling people to use the edge channel.

What do you think?

ohemorange:
> I think we should go ahead and update the instructions to use the beta channel now. Our DNS plugin snaps have been in the beta channel since our last release and it avoids us ever telling people to use the edge channel.

Oh, great call.

_scripts/instruction-widget/templates/install/snap.html
```diff
@@ -33,5 +30,45 @@ <h3>Snap Support</h3>
...
+<li>
+  Connect the DNS plugin snap to the Certbot snap
+  <p>
+    Run the following command, replacing &lt;PLUGIN&gt; with the name of your DNS provider.
+  </p>
+  <pre>sudo snap connect certbot:plugin {{dns_package_prefix}}-&lt;PLUGIN&gt;</pre>
+  <p>
+    For example, if your DNS provider is Cloudflare, you'd run the following command:
+  </p>
+  <pre>sudo snap connect certbot:plugin {{dns_package_prefix}}-cloudflare</pre>
```
bmw, line 71: This and the more general example above generates commands like:
```
sudo snap connect certbot:plugin --beta certbot-dns-cloudflare
```
which for me results in:
```
error: unknown flag `beta'
```

ohemorange: the solution I went with isn't the cleanest thing I've ever written, but I think it's better than the alternatives, which I saw as:
- hard-code the prefix of the plugin in `snap.html`
- refactor things again to deal with `--beta` being required for certbot install only on the wildcard tab
- hard-code install instructions for snaps

ohemorange:
Screenshot from 2020-08-18 12-55-27

_scripts/instruction-widget/templates/install/snap.html
bmw: Especially since this should just be temporary until we have stable DNS plugin snaps, I agree it's OK.

PR #615, opened 2020-07-24
Add a note about snap multiarch support

adferrand:
This PR creates a note to advert about snap multiarch support. It is associated to https://github.com/certbot/certbot/issues/8051.

### DIFF
--- a/_scripts/instruction-widget/templates/install/snap.html
+++ b/_scripts/instruction-widget/templates/install/snap.html
@@ -5,16 +5,8 @@
     <h3>Snap Support</h3>
   </div>
   <p>
-  Certbot snap support is currently in its beta phase. You can find instructions for
-  installing Certbot without using snap by selecting your OS in the dropdown above.
-  </p>
-</aside>
-<aside class="note">
-  <div class="note-header">
-    <h3>Snap Support</h3>
-  </div>
-  <p>
-  Certbot snap is available for the x86_64, ARMv7 and ARMv8 architectures.
+  Certbot snap support is currently in its beta phase and supports the x86_64, ARMv7, and ARMv8 architectures.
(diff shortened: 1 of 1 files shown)


_scripts/instruction-widget/templates/install/snap.html
```diff
@@ -5,9 +5,16 @@
...
+  Certbot snap support is currently in its beta phase. You can find instructions for
+  installing Certbot without using snap by selecting your OS in the dropdown above.
+  </p>
+</aside>
+<aside class="note">
+  <div class="note-header">
+    <h3>Snap Support</h3>
+  </div>
+  <p>
+  Certbot snap is available for the x86_64, ARMv7 and ARMv8 architectures.
```
bmw, line 17: nit: Rather than having two notes back to back with the header "Snap support", what do you think about just moving this into the previous note which could say something like:
```
Certbot snap support is currently in its beta phase and supports the x86_64, ARMv7, and ARMv8 architectures.
```

adferrand: This was my first approach, but then the declaration about 3 architectures supported sounded like a limitation, because the overall goal of this note is about warning users about limitations of the beta phase. By the way the note should have the color code of a warning here.

So I prefered writing it in a separate note to put some value on it.

bmw: I think it is a limitation that our other installation methods don't have. I also think it works nicely as a single note because it tells you what to do if the snap's beta status or supported architecture doesn't work for you without duplicating this information in each note. Finally, I think the back to back notes doesn't look great:
Screen Shot 2020-07-27 at 11 49 20 AM

adferrand: Done.

Issue #621, opened 2020-07-27
docker-compose instructions don't work


bmw:
The Docker instructions [here](https://github.com/certbot/website/tree/20d21a48ffd120c089f5c937321f19a73cfa65d8#with-docker) don't work. When I try to access `/instructions` or any of the specific pages of generated instructions such as `/lets-encrypt/snap-other` I get a 404.

Another problem we have now is the Dockerfile is still based on the old `node:8` image which is based on Debian 9. The newest version of Python Debian 9 has is Python 3.5 so we need to update to a newer image to be able to build our docs.

bmw:
In https://github.com/certbot/website/pull/622 I proposed removing these instructions until this is resolved.

bmw:
Fixed in #678.

PR #622, opened 2020-07-27
Remove broken Docker instructions

bmw:
I think we should remove the instructions until https://github.com/certbot/website/issues/621 is resolved.


ohemorange:
There are merge conflicts here

Issue #611, opened 2020-07-22
Stop using certbot-auto in _docs.sh


bmw:
We use certbot-auto [here](https://github.com/certbot/website/blob/3ebf9afce6c2cea60498b0a4c6e6abe918920173/_docs.sh#L7) to install OS dependencies, but I think we should instead use commands like the `apt` commands [here](https://certbot.eff.org/docs/contributing.html#running-a-local-copy-of-the-client). As part of this, I think we should switch the venv installation to Python 3.

These changes are to allow the deprecation of certbot-auto and Python 2 support in the future.

bmw:
Fixed in https://github.com/certbot/website/pull/628.

PR #628, opened 2020-08-13
Switch to Python 3 and stop using certbot-auto

bmw:
Fixes https://github.com/certbot/certbot/issues/8164 and https://github.com/certbot/website/issues/611.

The addition of `sudo` in the Dockerfile was needed because `certbot-auto` was previously handling calling `sudo` or not based on whether or not we were already root.

Both the Dockerfile and the website build with this change. The only change I could find is new warnings at build time when trying to document `certbot.compat.os`, however, the resulting files appear to be the same (probably except any differences between Python 2 and Python 3 stdlib `os`). I fixed the warnings in https://github.com/certbot/certbot/pull/8209.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,6 +24,7 @@ RUN apt-get install -y --no-install-recommends \
     gsfonts \
     latexmk \
     rsync \
+    sudo \
     texlive \
     texlive-latex-extra
 
--- a/_docs.sh
+++ b/_docs.sh
@@ -4,9 +4,9 @@ case "$1" in
   "depend" )
     cd _docs
     export CERTBOT_WEBSITE='True'
-    sudo apt update
-    sudo apt install python3-dev python3-venv gcc libaugeas0 libssl-dev \
-                     libffi-dev ca-certificates openssl -y
+    sudo apt-get update
+    sudo apt-get install python3-dev python3-venv gcc libaugeas0 libssl-dev \
+                         libffi-dev ca-certificates openssl -y
     ./tools/venv3.py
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -24,6 +24,7 @@ RUN apt-get install -y --no-install-recommends \
     gsfonts \
     latexmk \
     rsync \
+    sudo \
```
ohemorange, line 27: How was it working without this? Wasn't certbot-auto calling sudo internally? Why didn't it need to be explicitly installed before?

bmw: I tried to explain this in my PR description where I said:

> The addition of `sudo` in the Dockerfile was needed because `certbot-auto` was previously handling calling `sudo` or not based on whether or not we were already root.

To expand on that though, [certbot-auto only sets up sudo/su to be used if it is needed](https://github.com/certbot/certbot/blob/2d62dec7ecd7f62a77768aa951342b5cbfc06abc/certbot-auto#L189-L195), [it's needed in Travis](https://docs.travis-ci.com/user/installing-dependencies/#installing-packages-on-standard-infrastructure), but technically not needed in the Dockerfile because the build is running as root inside the container. Since `certbot-auto` was already running as `root` in the Docker image, it didn't call `su` or `sudo`.

I added the dependency on `sudo` to the Dockerfile so we could use the same command in both cases for simplicity.

ohemorange: Ooooh, I see now, I didn't realize it was running in two different environments, that makes sense.

PR #632, opened 2020-08-24
adding "return" to instructions function

### LINKED ISSUE #626: Runtime Mustache rendering is causing raw Jekyll markup to appear
While working on #612 I've been seeing some pretty ugly <a href="https://en.wikipedia.org/wiki/Flash_of_unstyled_content">FOUC</a> (well, more like Flash of Template Code) when I select a webserver/distro.

[Video example](https://tools.letsdebug.net/certbot-jekyll-fouc.mp4).

Two things seem to be happening:

- Upon input selection, `document.location.href` is changed and the browser begins navigating to the new URL via full page reload
- While that navigation happens, the mustache template is *also* being rendered. That template's output contains raw Jekyll markup such as:

```html
<div class="tab advanced">wildcard{% include tooltip.html term-name="wildcard-certificate" text="*" %}</div>
```

which you can see being rendered in the video.

Some time later, depending how fast the network is, the full page navigation completes and the unwanted markup disappears - since the static version of the page is properly rendered.

In my opinion, removing runtime mustache rendering entirely would be no big loss. I had a quick look around and I don't think the site actually relies on it in any way that can't be reasonably substituted with full navigation.

Otherwise, let's fix the mustache templates so they are not spitting out Jekyll markup.

I understand that perhaps not the most time can be afforded to this repo, but since this could be fixed by deleting code only, maybe it's worth it.

lschatzkin:
Fixes #626.

### DIFF
--- a/_scripts/instruction-widget/main.js
+++ b/_scripts/instruction-widget/main.js
@@ -107,6 +107,7 @@ InstructionWidget = (function() {
         if (input.os && input.webserver) {
           // We're on the homepage, redirect to instructions page
           window.location.href = instruction_url(input.os,input.webserver);
+          return;
         }
       }
       Instructions().render(content_container, input);


ohemorange:
I mean, I definitely believe you, but also what is going on here? Was this causing a bug?

PR #634, opened 2020-09-02
Changing  Veerotech from partial to full and added links

linuxwizard:
Adding Veerotech to full category as they support Lets Encrypt certificate for domains in all category of hosting

### DIFF
--- a/_data/hosting_providers.json
+++ b/_data/hosting_providers.json
@@ -1851,12 +1851,12 @@
   },
   {
     "name": "VeeroTech",
-    "link": "",
-    "category": "partial",
+    "link": "https://www.veerotech.net",
+    "category": "full",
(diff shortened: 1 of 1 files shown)


bmw:
Our definition of "full" support from https://certbot.eff.org/hosting_providers/ is:

> Hosting providers that offer free Let’s Encrypt HTTPS certificates automatically for their shared hosting product. No need to do anything on your end if you use their service.

while "partial" says:

> Hosting providers that offer free Let’s Encrypt HTTPS certificates for their shared hosting product, but require configuration. You’ll need to follow a tutorial.

The announcement you linked describes the steps that users must manually take to set up HTTPS on their site so I think VeeroTech should still be listed as partial support. Because of this, I'm closing this PR, but if I have misunderstood something, please let me know and I'll consider reopening.

PR #639, opened 2020-09-08
Updated HostPapa

Suprnova123:
Updated support for Let's Encrypt, changed name to be stylized as it is on the website, TLD shouldn't be included since the page will redirect to the appropriate one based on location anyways.


ohemorange (approved):
Hey awesome, thanks for supporting certificates automatically!

PR #640, opened 2020-09-10
Add a note to help snap users configure PATH under all shells and sudo

adferrand:
This PR adds a note to help users who hit the issue described in certbot/certbot#8246, providing convenient shell commands to add `/snap/bin` to `PATH` with zsh and fish.

Ultimately this issue should be fixed on snapd side (see https://bugs.launchpad.net/ubuntu/+source/zsh/+bug/1640514) but it the mean time I hope this will help users who want to use the Certbot snap with other shells than bash.

### DIFF
--- a/_scripts/instruction-widget/templates/install/snap.html
+++ b/_scripts/instruction-widget/templates/install/snap.html
@@ -35,6 +35,7 @@
 {{#advanced}}
 {{#dns_plugins}}
 {{>installcertbotwildcard}}
+{{>preparecertbotsnapcommand}}
 <li>
   Confirm plugin containment level
   <p>
@@ -53,11 +54,7 @@
 
 {{^advanced}}
 {{>installcertbot}}
-  <p>
-  Execute the following instruction on the command line on the machine to ensure
-  that the <tt>certbot</tt> command can be run.
-  </p>
-  <pre>sudo ln -s /snap/bin/certbot /usr/bin/certbot</pre>
+{{>preparecertbotsnapcommand}}
 {{/advanced}}
 
 {{> dnspluginssetup}}
--- a/_scripts/instruction-widget/templates/install/snap.html
+++ b/_scripts/instruction-widget/templates/install/snap.html
@@ -53,6 +53,11 @@
 
 {{^advanced}}
 {{>installcertbot}}
+  <p>
+  Execute the following instruction on the command line on the machine to ensure
+  that <tt>certbot</tt> command can be run.
+  </p>
+  <pre>sudo ln -s /snap/bin/certbot /usr/sbin/certbot</pre>
 {{/advanced}}
 
 {{> dnspluginssetup}}
--- a/_scripts/instruction-widget/templates/install/snap.html
+++ b/_scripts/instruction-widget/templates/install/snap.html
@@ -55,7 +55,7 @@
 {{>installcertbot}}
   <p>
   Execute the following instruction on the command line on the machine to ensure
-  that <tt>certbot</tt> command can be run.
+  that the <tt>certbot</tt> command can be run.
   </p>
   <pre>sudo ln -s /snap/bin/certbot /usr/bin/certbot</pre>
 {{/advanced}}
--- a/_scripts/instruction-widget/templates/install/snap.html
+++ b/_scripts/instruction-widget/templates/install/snap.html
@@ -17,6 +17,25 @@
     <p class="centered">
     <a class="link-button" href='https://snapcraft.io/docs/installing-snapd/'>install snapd</a>
     </p>
+
+    <aside class="note">
+      <div class="note-header">
+        <h3>Configuring snapd on specific shells</h3>
+      </div>
+      <p>
+      On shells like <tt>zsh</tt> or <tt>fish</tt>, the folder <tt>/snap/bin</tt>, which contains the snaps installed on your system,
+      needs to be added to the <tt>PATH</tt> environment variable. You will need to execute one of the following command in your
+      shell to add it permanently.<br/>
+      </p>
+      <p>
+        For <tt>zsh</tt>:
+        <pre style="margin:0;">echo -n 'export PATH=$PATH:/snap/bin' >> ~/.zshrc && . ~/.zshrc</pre>
+      </p>
+      <p>
+        For <tt>fish</tt>:
+        <pre style="margin:0;">echo -n 'set -gx PATH $PATH /snap/bin' >> ~/.config/fish/config.fish && . ~/.config/fish/config.fish</pre>
+      </p>
+    </aside>
 </li>
 <li>
     Remove any Certbot OS packages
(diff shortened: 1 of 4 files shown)


bmw (changes requested):
While testing this I found a different issue which I think may affect how we want to handle things here.

I was testing on CentOS 7 using `zsh` and after modifying `~/.zshrc` I tried running `sudo certbot ...` per our instructions and got:
```
sudo: certbot: command not found
```
The issue here is `certbot` isn't in the list of paths trusted by `sudo`. I googled around about this a bit and found https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/1595558 which Ubuntu chose to fix by patching the Ubuntu `sudo` package rather than making a change in `snapd` itself.

Do we also want to add instructions to have users modify their `sudo` config? Do we want to add instructions like the ones in this PR and then change all of our Certbot commands to something like `sudo -i certbot ...`?

Alternatively, we could skip both of these things and tell users to run either `sudo /snap/bin/certbot ...` or `sudo snap run certbot ...`.

I'm curious what you think, but I'm honestly just leaning towards changing all of our instructions to have people run `sudo /snap/bin/certbot`. It's similar to what we did in our `certbot-auto` instructions where we tell people to run `sudo /usr/local/bin/certbot-auto`.

If we want to continue with the approach in this PR though, the two nitpicky things I noticed so far are:

1. I think we should remove `-n` from the `echo` commands so the config file ends in a newline.
2. I wonder if we shouldn't modify the global config files for fish and zsh to ensure the change is made for all users on the system.

alexzorin:
>I'm honestly just leaning towards changing all of our instructions to have people run sudo /snap/bin/certbot. It's similar to what we did in our certbot-auto instructions where we tell people to run sudo /usr/local/bin/certbot-auto.

I think we should really try to avoid creating yet another unique way to invoke Certbot (certbot, /path/to/certbot-auto, letsencrypt, /path/to/letsencrypt-auto, /path/to/venv/bin/certbot, docker run certbot). It's hard to remember, especially if months have elapsed since last time you had to use it, or if you're just straight up a different person to whoever set it up initially.

How about asking the user to just symlink it into the `secure_path`:

    sudo ln -s /snap/bin/certbot /usr/sbin/certbot

adferrand:
Indeed. I think that using the full path to call Certbot is not very elegant, and could demotivate some people to use snaps.

If invoking a command to manipulate the `PATH` is not working consistently accross Linux distributions, I am more in favor in using the symlink approach proposed by @alexzorin. I would prefer however to put the symlink in `/usr/bin` instead of `/usr/sbin` in this case, since Certbot does not technically require root to run, and could be used as normal user (with several tweaks, arguably ^^).

Also in this case, I think we could simplify the documentation here by asking users to run this command unconditionnally (so independently from the actual shell used).

bmw:
I agree that using a symlink is a better idea. Thanks for proposing it Alex.

As for `/usr/bin/certbot` vs. `/usr/sbin/certbot`, I don't have a strong opinion but to share a few thoughts:

1. Both Debian and EPEL put Certbot in `/usr/bin/certbot`. On the one hand maybe this means we should do the same but on the other using `/usr/sbin` avoids potential conflicts.
2. I think you could make a decent argument for either location. It's true that you don't always need `root` with Certbot, however, it defaults to needing it and it is strictly needed for much of its functionality on Linux such as modifying `/var/www/html`, your Apache or Nginx config, or binding to port 80. I'll also flag that server software like `apache2` and `nginx` also don't _need_ `root` to run but are in `/usr/sbin` on both Debian and RHEL.

I'm personally fine with either location.

alexzorin:
I think `/usr/bin` is good because it matches the existing package locations, which also makes it a good canary for problems: if something exists at that path already, it means the user has not fully removed their previous packages, and `ln` will error out.

I just took a random path out of `secure_path`, no big intention behind it.

adferrand:
So @bmw, @alexzorin, are you agree to switch this PR content to the addition of a step after snapd has been installed to run `sudo ln -s /snap/bin/certbot /usr/sbin/certbot` unconditionally, instead of specific `PATH` manipulations for zsh and fish shells?

bmw:
I think we should do it after the Certbot snap is installed and maybe use `/usr/bin` per Alex's suggestion unless you now have a strong preference for `/usr/sbin`, but otherwise, this seems like the best plan to me!

adferrand:
Yes sorry, I meant after certbot install, not snapd of course. Otherwise I still have a slight preference for `/usr/bin` which seems more decent for a not system oriented binary (despite apache/nginx).

adferrand:
I update this PR to use the new comment as discussed.

_scripts/instruction-widget/templates/install/snap.html
```diff
@@ -72,6 +53,11 @@ <h3>Configuring snapd on specific shells</h3>
 
 {{^advanced}}
 {{>installcertbot}}
+  <p>
+  Execute the following instruction on the command line on the machine to ensure
+  that <tt>certbot</tt> command can be run.
+  </p>
+  <pre>sudo ln -s /snap/bin/certbot /usr/sbin/certbot</pre>
```
alexzorin, line 60: The conversation is a bit confused, but I think we agreed on `/usr/bin/` in the end?

adferrand: Indeed!

_scripts/instruction-widget/templates/install/snap.html
```diff
@@ -53,6 +53,11 @@ <h3>Snap Support</h3>
 
 {{^advanced}}
 {{>installcertbot}}
+  <p>
+  Execute the following instruction on the command line on the machine to ensure
+  that <tt>certbot</tt> command can be run.
```
bmw, line 58:
```suggestion
  that the <tt>certbot</tt> command can be run.
```

_scripts/instruction-widget/templates/install/snap.html
```diff
@@ -53,6 +53,11 @@ <h3>Snap Support</h3>
 
 {{^advanced}}
```
bmw, line 54: Since this is enclosed in `{{^advanced}}` it doesn't show up in our wildcard instructions which I think should be fixed.

_scripts/instruction-widget/templates/install/snap.html
```diff
@@ -53,6 +53,11 @@ <h3>Snap Support</h3>
...
 {{>installcertbot}}
+  <p>
```
bmw, line 56: nit: I personally think it'd be a good idea to make this a separate numbered released instruction by enclosing it with `<li>` tags. I worry that without it, people might miss it.

bmw (changes requested):

_scripts/instruction-widget/templates/install/snap.html
adferrand: You are right. I put it in a dedicated step.

_scripts/instruction-widget/templates/install/snap.html
adferrand: Fixed.

bmw (changes requested):
I just have the same grammar correction that got lost in the refactor but this otherwise LGTM!

bmw (approved):

alexzorin:
Does this appear on the site anywhere right now?

bmw:
It doesn't. It's supposed to, but the automatic deployments to the live site are currently broken.

EFF's tech ops team is working on it.

PR #642, opened 2020-09-12
Update hosting_providers.json

reupke:
Updated listing for Netsite.


ohemorange (approved):
Nice, thanks for supporting automatic https!

PR #647, opened 2020-09-22
Add Firebase Hosting to providers

### LINKED ISSUE #646: Add Firebase Hosting to hosting providers
Firebase hosting [offers automatic, zero-configuration](https://firebase.google.com/products/hosting) SSL for all sites and domains. It technically isn't shared hosting, but neither is NFSN, which is on the list

derekrobot:
Fixes #646


ohemorange (approved):
Looks good, thanks!

PR #650, opened 2020-09-28
Update snap instructions to ensure snapd is up to date

adferrand:
Partial fix for #8280

Similarly with what #8313 directly instruct users in case of certbot snap failure, this PR adds the relevant upgrade command for `snapd` in the certbot installation instruction in snap mode.

### DIFF
--- a/_scripts/instruction-widget/templates/install/snap.html
+++ b/_scripts/instruction-widget/templates/install/snap.html
@@ -19,6 +19,14 @@
     </p>
 </li>
 <li>
+  Ensure that your version of snapd is up to date
+  <p>
+  Execute the following instruction on the command line on the machine to ensure
+  that you have the latest version of <tt>snapd</tt>.
+  </p>
+  <pre>sudo snap install core; sudo snap refresh</pre>
+</li>
+<li>
     Remove any Certbot OS packages
     <p>
     If you have any Certbot packages installed using an OS package manager like
--- a/_scripts/instruction-widget/templates/install/snap.html
+++ b/_scripts/instruction-widget/templates/install/snap.html
@@ -24,7 +24,7 @@
   Execute the following instructions on the command line on the machine to ensure
   that you have the latest version of <tt>snapd</tt>.
   </p>
-  <pre>sudo snap install core; sudo snap refresh</pre>
+  <pre>sudo snap install core; sudo snap refresh core</pre>
 </li>
 <li>
     Remove any Certbot OS packages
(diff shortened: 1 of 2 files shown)


bmw:
Don't we want to do this before the user tries to install the Certbot snap? At least on Debian 9 if you try to install the Certbot snap it fails with:
```
error: cannot install "certbot": snap not found
```
To fix this, I'd put this in `_scripts/instruction-widget/templates/install/snap.html` earlier in the file.

adferrand:
You are right, it is sufficient for Debian 10 but not Debian 9.

_scripts/instruction-widget/templates/install/snap.html
```diff
@@ -18,6 +18,14 @@ <h3>Snap Support</h3>
     <a class="link-button" href='https://snapcraft.io/docs/installing-snapd/'>install snapd</a>
     </p>
 </li>
+<li>
+  Ensure that your version of snapd is up to date
+  <p>
+  Execute the following instruction on the command line on the machine to ensure
+  that you have the latest version of <tt>snapd</tt>.
+  </p>
+  <pre>sudo snap install core; sudo snap refresh</pre>
```
bmw, line 27:
```suggestion
  <pre>sudo snap install core; sudo snap refresh core</pre>
```
This is definitely a nitpick, but I think we should just refresh the core snap instead of taking the time to try and refresh every snap on their system.

I double checked that this syntax works on Debian 9 even before the `core` snap is installed and it does.

_scripts/instruction-widget/templates/install/snap.html
```diff
@@ -18,6 +18,14 @@ <h3>Snap Support</h3>
     <a class="link-button" href='https://snapcraft.io/docs/installing-snapd/'>install snapd</a>
     </p>
 </li>
+<li>
+  Ensure that your version of snapd is up to date
+  <p>
+  Execute the following instruction on the command line on the machine to ensure
```
bmw, line 24:
```suggestion
  Execute the following instructions on the command line on the machine to ensure
```

bmw (changes requested):
Couple nitpicks but this otherwise LGTM!

_scripts/instruction-widget/templates/install/snap.html
adferrand: Alright. In this case I also update https://github.com/certbot/certbot/pull/8313 to keep the instructions consistent between the website and certbot.

bmw: Thanks! I personally think that makes sense.

bmw (approved):

PR #658, opened 2020-10-07
Update Debian instructions to snaps

bmw:
Fixes https://github.com/certbot/certbot/issues/8293.

I tested this following the instructions at https://github.com/certbot/website#building-with-travis and confirmed that the Debian instructions now use the snaps while the Devuan instructions which cannot use snaps still have installation instructions using `apt-get`.

### DIFF
--- a/_scripts/instruction-widget/install.js
+++ b/_scripts/instruction-widget/install.js
@@ -39,9 +39,6 @@ module.exports = function(context) {
     else if (snap_distros.includes(context.distro)) {
       snap_install();
     }
-    else if (context.distro == "debian" && context.version > 8) {
-      debian_install();
-    }
     // @todo: Implement or complete these.
     // else if (context.distro == "python"){
     //   return pip_install();
--- a/_scripts/instruction-widget/install.js
+++ b/_scripts/instruction-widget/install.js
@@ -28,7 +28,7 @@ module.exports = function(context) {
 
     // This is the list of distributions that should be shown our snap
     // instructions.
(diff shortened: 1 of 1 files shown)


_scripts/instruction-widget/install.js
```diff
@@ -28,7 +28,7 @@ module.exports = function(context) {
 
     // This is the list of distributions that should be shown our snap
     // instructions.
-    snap_distros = ["snap", "ubuntu", "arch", "opensuse", "fedora"];
+    snap_distros = ["snap", "ubuntu", "arch", "opensuse", "fedora", "debian"];
```
adferrand, line 31: By the way, I missed that in the previous PR. Could you add the `var` keyword on this variable declaration, to avoid it to be global?

adferrand (changes requested):
Aside the inline command, can you add the comment about `debian_install` being unused, like it is done for the other functions that are on the same situation.

Bonus point to add the missing `var` on all install functions  + `html` ^^

_scripts/instruction-widget/install.js
bmw: I don't love blocking PRs on fixing problems the PR didn't introduce, but this is done.

bmw:
`debian_install` is still being used for Devuan.

adferrand:
Sorry I missed that.

bmw:
Testing things again following https://github.com/certbot/website#building-with-travis, it all still seems to work to me.

adferrand (approved):
LGTM!

Issue #619, opened 2020-07-27
Explaining supported architectures in snap instructions


bmw:
The Certbot snap which we plan to make our primary installation method supports the x86_64, ARMv7, and ARMv8 CPU architectures.

According to the [statistics](https://gist.github.com/bmw/7970ee3e3f4c8f8bad8e193a646c6c9b) I pulled from PyPI (which counts Certbot installations using `pip` including those from `certbot-auto`), this covers 97.8% of our user base. I also suspect this situation will get better over time as the older architectures continue to become less popular and supported.

The question I have is how should we approach this issue in Certbot's instruction generator? Should we make a nice glossary/tooltip item about this and explicitly list the supported architectures? I worry a bit about this bogging down the installation process and confusing novice users, most of whom won't have to worry about this, but that probably seems like the best approach to me personally.

bmw:
For what it's worth, I pulled some new stats here which can be found at https://gist.github.com/bmw/c11817038a71f25609eb8f003f7c09e2.

ohemorange:
Can snap itself be installed on the other architectures? The first step is to install `snapd`; if snapcraft's messages about why that doesn't work are sufficient, we may not need to worry about it.

bmw:
>Can snap itself be installed on the other architectures?

Yes. For example, https://packages.debian.org/search?keywords=snapd&searchon=names&suite=stable&section=all shows Debian stable having `snapd` packaged for `amd64 arm64 armel armhf i386 ppc64el s390x`.

bmw:
The error you get from the `snap install` command if you do install `snapd` though is:
```
# snap install certbot-dns-cloudflare
error: snap "certbot-dns-cloudflare" is not available on this architecture (i386) but exists on
       other architectures (amd64, arm64, armhf).
```
(The error is different on Certbot due to https://forum.snapcraft.io/t/is-it-possible-to-drop-an-unsupported-architecture-from-the-store/10510. That output is:
```
error: snap "certbot" is not available on stable but is available to install on the following
       channels:

       edge       snap install --edge certbot

       Please be mindful pre-release channels may include features not completely tested or
       implemented. Get more information with 'snap info certbot'.
```
)

bmw:
To be clear, I don't have a strong opinion on whether or not we have a note. With that said, to state my preference, I suppose I slightly prefer saying something but at least in the long term documenting it very well like we do all of our other confusing terminology.

Maybe adding it here in the generated snap instructions makes the most sense?
[...]

PR #659, opened 2020-10-07
Instruct RHEL to use snaps when possible

bmw:
Fixes https://github.com/certbot/certbot/issues/8295 with the exception of removing instructions for CentOS/RHEL 6. If this PR is merged, I will create an issue to track that piece.

In my initial plan at https://docs.google.com/document/d/174ILmAfVlqWvYH9n8gkQnLc-FECiIsKpa2nZCxEklik/edit?usp=sharing, I suggested waiting to do this, but I think we may as well do this now. It should help our snaps be more widely tested before we deprecate certbot-auto on the platform and if something goes wrong, we can always revert this PR.

I tested this following the instructions at https://github.com/certbot/website#building-with-travis and confirmed that the CentOS/RHEL 7+ and Fedora instructions now use the snaps while the CentOS/RHEL 6 instructions still direct people to `certbot-auto`.

### DIFF
--- a/_scripts/instruction-widget/install.js
+++ b/_scripts/instruction-widget/install.js
@@ -52,9 +52,6 @@ module.exports = function(context) {
     else if ((context.distro == "opbsd")||(context.distro =="freebsd")){
       bsd_install();
     }
-    else if (context.distro == "fedora"){
-      fedora_install();
-    }
     else if (context.distro == "centos" || context.distro == "rhel") {
       // The oldest version of RHEL where snapd is packaged is RHEL 7.
       if (context.version < 7) {
--- a/_scripts/instruction-widget/install.js
+++ b/_scripts/instruction-widget/install.js
@@ -28,7 +28,7 @@ module.exports = function(context) {
 
     // This is the list of distributions that should be shown our snap
     // instructions.
-    snap_distros = ["snap", "ubuntu", "arch", "opensuse"];
+    snap_distros = ["snap", "ubuntu", "arch", "opensuse", "fedora"];
 
     // Each case listed here should map to a template.
     // They don't necessarily need to map to distros.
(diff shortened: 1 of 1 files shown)


nbebout:
@bmw Does this mean you will no longer tell people about our packages in the Fedora and EPEL repos?

_scripts/instruction-widget/install.js
```diff
@@ -194,6 +199,8 @@ module.exports = function(context) {
     context.dns_package_prefix = "certbot-dns";
   }
 
+  // This function is currently unused, but we keep it around to make it easy
```
adferrand, line 202: Have you a specific short/middle-term use case in mind to keep the function around? Otherwise I think that the git history may be sufficient here.

adferrand (changes requested):
Aside my inline comment, the branch testing `context.distro == "fedora"` can be removed.

_scripts/instruction-widget/install.js
bmw: I think if we find a significant problem with our snaps on these systems, we'd want to change things back.

Another case is we want instructions for other architectures as suggested [here](https://github.com/certbot/website/issues/619#issuecomment-668205864).

I could clean some of this dead code up, but this PR follows the same pattern as https://github.com/certbot/website/pull/627 and https://github.com/certbot/website/pull/648 and I'd personally like to continue it and we can clean everything up later when we know we don't need it.

bmw: Ah I just noticed I did clean up the unused branches in those PRs so I'm happy to do that here. I'd still like to leave this function around though.

adferrand: That works for me. With the branch removed, it becomes clear from any decent IDE that the function is unused and can be deleted during future refactorings.

adferrand (approved):
LGTM!

Indeed @nbebout, it means that we do not advertise the existence of the OS packages on the official website for environments where snap is officially supported. It was also the case for `certbot-auto` before, and it constitutes an effort to align as much as possible the distribution channel for Certbot on our end.

bmw:
Thanks for the review.

The issue to remove the CentOS/RHEL 6 instructions is https://github.com/certbot/certbot/issues/8358.

PR #661, opened 2020-10-07
Add extra explanation to wildcard glossary about base domain

### LINKED ISSUE #660: Current wildcard glossary not clear to users about the base domain
https://community.letsencrypt.org/t/combining-wildcard-and-regular-certificates/135502 demonstrates the lack of certain information of the current wildcard glossary: the example says the wildcard certificate isn't valid for the base domain.

While this is true, the topic demonstrates that the glossary *might* tell you it's impossible to get a wildcard for the base domain, point. Personally, I think adding a little line with something like "For the wildcard certificate to be valid for `example.com` *too*, you will need to add the hostname `example.com` to the certificate *along side of* the wildcard hostname `*.example.com`."

osirisinferi:
Fixes #660


alexzorin (dismissed):
Big agree on this change, thanks.

Would you be interested in rewriting the first sentence as well? I have a lot of trouble myself understanding what it says.

If not, I'm happy to just merge this.

alexzorin (changes requested):
These are great - they all come up frequently.

1. Annoyingly, because this is markdown, many of the asterisks are being interpreted as *emphasis*. Could you please escape each of them as `\*`?
2. Let's split these explanations into 4 or 5 paragraphs

osirisinferi:
I'll try to learn Markdown in that case first.

alexzorin:
If you want a convenient way to see how it will be rendered (without actually starting this project up), the forums also use markdown. 

So you can just paste the text there and the forum post preview will faithfully render it, including the magic asterisks.

osirisinferi:
Also reworded some parts a little bit to make it more clear (I think).

alexzorin (changes requested):
Just some small nits and simplifications.

My suggested change of "Hostname"→"Name" might confuse novice readers, I'm not sure. But the idea of a wildcard hostname didn't feel good to me.

alexzorin (approved):
Thanks a lot, I think this is a huge improvement.

One of the wildcard tooltips has messed up formatting, but since that was already an issue before, let's fix that separately.

PR #663, opened 2020-10-08
Add Easyname to hoster list

deadda7a:
I was not sure if I should use "full" or "partial" as a category - You have to take steps to enable TLS, but it's only flipping a switch.


alexzorin:
Hello @deadda7a,

Thank for for your submission. It appears that easyname [is already listed](https://github.com/certbot/website/blob/7e59c145f198b3121c9e797295ea515783aef0bd/_data/hosting_providers.json#L2262-L2271) using their English language website.

I am going to close this as a duplicate, but if you had a particular reason to submit easyname again, please reply back.

deadda7a:
WTF I checked the list like 4 times. But to my defence, I was really tired.

PR #671, opened 2020-11-11
Update hosting_providers.json

BoryaGames:
I am added hosting 7m.pl

### DIFF
--- a/_data/hosting_providers.json
+++ b/_data/hosting_providers.json
@@ -2448,5 +2448,15 @@
     "plan": "",
     "reviewed": "2020.09.28",
     "note": "Let's Encrypt available on all hosting, VPS, and dedicated server plans."
+  },
+  {
+    "name": "7m pl",
+    "link": "https://7m.pl",
+    "category": "full",
+    "tutorial": "",
+    "announcement": "",
+    "plan": "",
+    "reviewed": "",
+    "note": ""
   }
 ]


_data/hosting_providers.json
```diff
@@ -2448,5 +2448,15 @@
...
     "note": "Let's Encrypt available on all hosting, VPS, and dedicated server plans."
+  },
+  {
+    "name": "7m pl",
+    "link": "https://7m.pl",
+    "category": "full",
+    "tutorial": "",
+    "announcement": "",
+    "plan": "",
+    "reviewed": "",
```
alexzorin, line 2459:
```suggestion
    "reviewed": "2020.11.12",
```

alexzorin (changes requested):
Hi @BoryaGames,

Could you please clarify whether the hosting provider provides automatic HTTPS for custom domains as well, not only for `*.7m.pl`? If so, are there any example domains?

BoryaGames:
> Hi @BoryaGames,
> 
> Could you please clarify whether the hosting provider provides automatic HTTPS for custom domains as well, not only for `*.7m.pl`? If so, are there any example domains?

Yes, but i am don't know example domains.

alexzorin:
I wasn't able to verify that HTTPS for custom domains is supported, so I sent the provider an email. I received a response:

>We don't currently offer https for domains, but it will be available soon.

I think the appropriate thing to do here would be to wait for this feature to be implemented.

PR #676, opened 2020-12-01
Remove instructions for CentOS 6 and RHEL 6

adferrand:
Fixes https://github.com/certbot/certbot/issues/8358

I followed the same approach than for previous instructions removals, by editing the minimal set of logic required to reach the target, but keeping existing methods and templates.

I think we should make a real clean of the website codebase eventually. If you agree, I can make a PR for it.

### DIFF
--- a/_data/inputs.json
+++ b/_data/inputs.json
@@ -99,18 +99,6 @@
             "version": "0"
         },
         {
-            "name": "CentOS 6",
-            "id": "centos6",
-            "distro": "centos",
-            "version": "6"
-        },
-        {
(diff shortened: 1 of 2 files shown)


bmw (approved):
Thanks for the PR!

I think there's a chance we'll want the instructions back for non-EOL'd distros for the next few months as we see how the transition off certbot-auto goes. With that in mind, I personally would recommend just creating an issue to track cleaning up the code for us to do when we're sure we don't need it. Alternatively, we could do the work now and revert later if needed, but I personally would prioritize other work before this.

PR #689, opened 2021-01-08
Add cloudflare to list of SSL hosting providers

lerichardson:


_data/hosting_providers.json
```diff
@@ -2478,5 +2478,14 @@
     "plan": "",
     "reviewed": "2020.09.28",
     "note": "Let's Encrypt available on all hosting, VPS, and dedicated server plans."
+  },
+  {
+    "name": "Cloudflare",
+    "link": "https://cloudflare.com",
+    "category": "full",
+    "tutorial": "https://developers.cloudflare.com/ssl/client-certificates",
```
bmw, line 2486: This link seems to be about client side TLS certificates which allows servers to authenticate connecting clients rather than server side TLS certificates which allow things like an average user to connect to a site over HTTPS rather than HTTP which is the focus of Certbot.

Assuming Cloudflare also offers server side certificates, is there a link to documentation about their offering there? I don't think us linking to documentation about client side certificates is correct.

alexzorin: I think if Cloudflare is to be listed, the listing should name and point to their hosting products, so it's clear that we're not talking about the main CDN product that most people are familiar with. 

IMO it would not be a benefit to dilute the "shared hosting product" list with CDNs.

What we could list:

- [Workers Sites](https://developers.cloudflare.com/workers/platform/sites) 
- [Cloudflare Pages](https://pages.cloudflare.com/) (not GA, "request beta access")

lerichardson:
I'm a bit confused. Cloudflare seems to offer several certificates. Which one(s) are you talking about?
Screenshot 2021-01-09 104859

_data/hosting_providers.json
lerichardson: I believe [this link](https://developers.cloudflare.com/ssl/universal-ssl) is correct instead, but I would like to have it checked before I edit the pull request.

bmw: I agree with Alex and think that adding CDNs to our hosting providers lists isn't really appropriate. Please correct me if I'm mistaken, but I believe that link refers to the TLS certificates you get when using Cloudflare as a CDN (and after doing that you still should set up TLS between your server and Cloudflare).

If this is correct, I personally think we should either:

1. Update the link, name, etc. to refer to Cloudflare Workers as Alex suggested.
2. Close this PR.

Unless you have a particular reason for wanting to list hosting Cloudflare Workers, I'm personally leaning towards the latter. I appreciate you taking the time to open a PR which resulted in a useful discussion about whether CDNs belong on this list, however, I think what Cloudflare offers isn't really an alternative to Let's Encrypt and Certbot.

PR #690, opened 2021-01-08
Add convenience script for reviewing branches from website-builds

### LINKED ISSUE #579: Simplify using Travis Builds
The instructions at https://github.com/certbot/website#travis-builds seemed to break during the website redesign as I initially reported at https://github.com/certbot/website/issues/443. https://github.com/certbot/website/pull/578 attempted to sort of make things work, but it's not pretty.

Is there anything we can do to simplify this? The best idea that initially springs to mind would be to provide a script/Docker setup in this repository that automatically fetches things a branch from the website-builds repository where the branch name is given as an environment variable or command line flag.

wioux:
Fixes #579. This adds a script `website-builds.sh` that takes a branch name as an argument, fetches the build of that branch from certbot/website-builds, and wraps it in an nginx docker container with nginx.conf.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,21 +10,17 @@ Website for [EFF's Certbot](https://certbot.eff.org/) project. Uses Jekyll for s
 
 If you're developing directly on this repository rather than on a fork, it's probably easiest to let Travis build the site for you.
 
-All branches and pull requests and built and tested by Travis.
+All branches and pull requests are built and tested by Travis.
 
 For branches, the built assets are pushed to an analagous branch in [certbot/website-builds](https://github.com/certbot/website-builds). Built assets from PRs are not saved because Travis doesn't provide a mechanism to securely push to a Github repo after PRs across forks.
 
-To view the build of any branch, checkout that branch from certbot/website-builds and run nginx to serve the files using the nginx configuration file from this certbot/website repository.
+To view the build of any branch, use the `website-builds.sh`, which will fetch the branch from certbot/website-builds and start a docker to serve the files.
 
-For example, commands to do this might looks like:
 ```
-git clone https://github.com/certbot/website-builds.git
-cd website-builds
-git checkout <RELEVANT BRANCH>
-CERTBOT_WEBSITE_PATH=/path/to/your/local/certbot/website/repo
-docker run -p 8000:4000 --rm -v "$CERTBOT_WEBSITE_PATH/nginx.conf:/etc/nginx/conf.d/default.conf:ro" -v $(pwd):/usr/share/nginx/html:ro -it nginx
+./website-builds.sh <NAME OF BRANCH>
 ```
-After starting that command running, you can access the website in your browser at http://localhost:8000. To shut the server down, just hit ctrl+c in the terminal you ran the docker command.
+
+After starting that command, you can access the website in your browser at http://localhost:8000. To shut the server down, just hit Ctrl+C in the terminal where `website-builds.sh` is running.
 
 If you are on linux and your user is not a member of the docker group, you'll need to run the command with `sudo`.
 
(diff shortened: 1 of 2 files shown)


ohemorange:
I tried restarting the build, not sure why we're hitting that error. Looking into it.

ohemorange:
Ah, well, that would explain it -- looks like #685 isn't integrated here. Maybe merging in master will fix this...

website-builds.sh
```diff
@@ -0,0 +1,30 @@
...
+FROM nginx:alpine
+
+RUN cd /usr/share/nginx \
+ && wget "https://github.com/certbot/website-builds/archive/$REF.zip" \
+ && unzip *.zip \
+ && mv website-builds-*/* html
+EOF
+
+echo
+echo "Starting server. You can access the website in your browser at http://localhost:8000. Press Ctrl-C to shut down."
```
ohemorange, line 28: this gets kind of lost since the docker/server output pretty much immediately kicks it off the screen, but I don't actually know if there's a good solution to that -- we probably actually do want that output visible.

wioux: I thought it'd be nice to put some output here anyway since the container doesn't emit any when it first comes up, to signal that the build has finished and the server is running. I think only the shutdown instruction is really lost, since if nginx logs kick the message off the screen that means the user is accessing http://localhost:8000 at least. I moved the note about Ctrl-C to a new line and added some escape codes to color the output... Does that work? Is it too much?

wioux:
> Ah, well, that would explain it -- looks like #685 isn't integrated here. Maybe merging in master will fix this...

Thanks!

website-builds.sh
ohemorange: I actually love the colored output, very official-looking. On mine the container is emitting stuff, that's the problem -- the last thing I get is `/docker-entrypoint.sh: Configuration complete; ready for start up`. I wonder if we can just pass that print statement to the docker container, so it prints it once it's done with startup? If not it's not the biggest deal, this is definitely much better, but the `localhost:8000` info is getting lost as well.
Screenshot from 2021-01-13 14-36-19

wioux: Ooh ok. Sorry, my nginx:alpine image was 21 months old! I pulled the latest and now I see the additional output. The image let's you silence the entrypoint messages with an environment variable, which I've set. How's it look to you now?

ohemorange: Ah, that explains it! Much better, thanks!!

README.md
```diff
@@ -10,21 +10,17 @@ Website for [EFF's Certbot](https://certbot.eff.org/) project. Uses Jekyll for s
 
 If you're developing directly on this repository rather than on a fork, it's probably easiest to let Travis build the site for you.
 
-All branches and pull requests and built and tested by Travis.
+All branches and pull requests are built and tested by Travis.
 
 For branches, the built assets are pushed to an analagous branch in [certbot/website-builds](https://github.com/certbot/website-builds). Built assets from PRs are not saved because Travis doesn't provide a mechanism to securely push to a Github repo after PRs across forks.
 
-To view the build of any branch, checkout that branch from certbot/website-builds and run nginx to serve the files using the nginx configuration file from this certbot/website repository.
+To view the build of any branch, use the `website-builds.sh`, which will fetch the branch from certbot/website-builds and start a docker to serve the files.
```
ohemorange, line 17:
```suggestion
To view the build of any branch, use the script `website-builds.sh`, which will fetch the branch from certbot/website-builds and start a docker to serve the files.
```

ohemorange (approved):
found one last typo, then looks good to go!

ohemorange (changes requested):
clicked the wrong button last time sorry

ohemorange (approved):
Thank you!!

PR #693, opened 2021-01-25
Add NDCHost.com to hosting providers list

ShaunR:
NDCHost.com is a privately owned and operated web hosting company that fired up it's first server in 2002.

### DIFF
--- a/_data/hosting_providers.json
+++ b/_data/hosting_providers.json
@@ -120,6 +120,16 @@
     "note": ""
   },
   {
+    "name": "NDCHost",
+    "link": "https://ndchost.com/",
+    "category": "full",
+    "tutorial": "",
+    "announcement": "",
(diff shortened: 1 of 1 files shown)


ohemorange:
Hello. For web hosting plans, do the users need to take any explicit steps to activate their free SSL certificate? If so, that's what we call "partial support," and in that case, please include a link to how to do in the "tutorial" section. If they are not required to do so, do you have a news release or splash page on your website that indicates that SSL is on by default?

ShaunR:
@ohemorange Free ssl is included and enabled by default on all of our web hosting plans so long as the user has not installed their own valid cert. Currently our website hosting plans does not show we include this but that page will be updated here in a few weeks.

ohemorange (approved):
thank you!

PR #698, opened 2021-02-25
Add pip instructions

ohemorange:
Questions:
- Should we give people the command to upgrade certbot monthly via cron and remove [Optional]?
- Should we switch to recommending people blow away the venv each time?
- Should the upgrade instruction be down below with renewal, so people can do the most important things first? It's a little more work (it'll need to be moved into a different partial) but super doable.
- We don't mention third-party plugins elsewhere; should we remove that from here?
- What should the system name in the uri and the dropdown/title be?

dropdown and heading:
Screenshot from 2021-02-25 15-19-36

default instructions:
Screenshot from 2021-02-25 15-17-18
Screenshot from 2021-02-25 15-17-34
Screenshot from 2021-02-25 15-17-44

wildcard instructions:
Screenshot from 2021-02-25 15-18-50

### DIFF
--- a/_scripts/instruction-widget/templates/getting-started/renewal.html
+++ b/_scripts/instruction-widget/templates/getting-started/renewal.html
@@ -73,3 +73,16 @@
   look for the lock icon in the URL bar.
   </p>
 </li>
+
+{{#upgrade_instructions}}
+<li>
+  [Optional] Upgrade certbot
+  <p>When you're ready to upgrade certbot, run the following command on the command line on the machine.
+    <pre class="no-before"><ol>
+        <li>{{install_command}} --upgrade {{package}}</li></ol></pre>
+    </p>
+
+  <p>If this step leads to errors, run <code>rm -rf /opt/certbot</code> and repeat all installation instructions.</p>
+  <p>For automatic upgrades, the command in this step can be run by a monthly cron job.</p>
+</li>
+{{/upgrade_instructions}}
--- a/_scripts/instruction-widget/templates/getting-started/renewal.html
+++ b/_scripts/instruction-widget/templates/getting-started/renewal.html
@@ -76,8 +76,9 @@
 
 {{#upgrade_instructions}}
 <li>
-  [Optional] Upgrade certbot
-  <p>When you're ready to upgrade certbot, run the following command on the command line on the machine.
+  [Monthly] Upgrade certbot
+  <p>It's important to occasionally update Certbot to keep it up-to-date. To do this, run the
+    following command on the command line on the machine.
     <pre class="no-before"><ol>
         <li>{{install_command}} --upgrade {{package}}</li></ol></pre>
     </p>
--- a/_scripts/instruction-widget/templates/getting-started/renewal.html
+++ b/_scripts/instruction-widget/templates/getting-started/renewal.html
@@ -84,6 +84,5 @@
     </p>
 
   <p>If this step leads to errors, run <code>rm -rf /opt/certbot</code> and repeat all installation instructions.</p>
-  <p>For automatic upgrades, the command in this step can be run by a monthly cron job.</p>
 </li>
 {{/upgrade_instructions}}
--- a/_scripts/instruction-widget/templates/getting-started/renewal.html
+++ b/_scripts/instruction-widget/templates/getting-started/renewal.html
@@ -80,9 +80,11 @@
   <p>It's important to occasionally update Certbot to keep it up-to-date. To do this, run the
     following command on the command line on the machine.
     <pre class="no-before"><ol>
-        <li>{{install_command}} --upgrade {{package}}</li></ol></pre>
+        <li>{{install_command}} --upgrade {{package}}{{#advanced}} certbot-dns-&lt;PLUGIN&gt;{{/advanced}}</li></ol></pre>
     </p>
 
-  <p>If this step leads to errors, run <code>rm -rf /opt/certbot</code> and repeat all installation instructions.</p>
+  {{#advanced}}<p>As above, replace <code>&lt;PLUGIN&gt;</code> with the name of your DNS provider.</p>{{/advanced}}
+
+  <p>If this step leads to errors, run <code>sudo rm -rf /opt/certbot</code> and repeat all installation instructions.</p>
 </li>
 {{/upgrade_instructions}}
--- a/_scripts/instruction-widget/templates/install/pip.html
+++ b/_scripts/instruction-widget/templates/install/pip.html
@@ -37,7 +37,7 @@
     "augeas on &lt;your_system_name&gt;" will probably yield helpful results.</p>
 </li>
 <li>
-  Set up a venv
+  Set up a Python virtual environment
   <p>
   Execute the following instructions on the command line on the machine to set up a virtual
   environment.
--- a/_scripts/instruction-widget/templates/install/pip.html
+++ b/_scripts/instruction-widget/templates/install/pip.html
@@ -13,8 +13,8 @@
 <li>
     Install system dependencies
 
-    <p>System dependencies may include Python 3.6+, including venv (this might be called <code>python3-dev</code>,
-      <code>python3-devel</code>, or <code>python3-venv</code>) and Augeas for the Apache
+    <p>System dependencies may include Python 3.6+, including venv (this might be called <code>python3</code>,
+      <code>python3</code>, or <code>python3-venv</code>) and Augeas for the Apache
       plugin (<code>libaugeas0</code> or <code>augeaslibs</code>).</p>
 
     <p>If you're having trouble installing cryptography, you may need to install additional
@@ -30,7 +30,7 @@
         <li># For RPM-based distributions (e.g. Fedora, CentOS ...)</li>
         <li># NB1: old distributions will use yum instead of dnf</li>
         <li># NB2: RHEL-based distributions use python3X instead of python3 (e.g. python36)</li>
-        <li>sudo dnf install python3-devel augeas-libs redhat-rpm-config</li></ol></pre>
+        <li>sudo dnf install python3 augeas-libs redhat-rpm-config</li></ol></pre>
     </p>
 
     <p>These packages may have slightly different names on other distributions; a search such as
--- a/_scripts/instruction-widget/templates/install/pip.html
+++ b/_scripts/instruction-widget/templates/install/pip.html
@@ -45,12 +45,12 @@
     Remove certbot-auto and any Certbot OS packages
     <p>
     If you have any Certbot packages installed using an OS package manager like
-    <tt>apt</tt>, <tt>dnf</tt>, or <tt>yum</tt>, you should remove them before
+    <code>apt</code>, <code>dnf</code>, or <code>yum</code>, you should remove them before
     installing the Certbot snap to ensure that when you run the command
-    <tt>certbot</tt> the snap is used rather than the installation from your OS
+    <code>certbot</code> the snap is used rather than the installation from your OS
     package manager. The exact command to do this depends on your OS, but
-    common examples are <tt>sudo apt-get remove certbot</tt>, <tt>sudo dnf
-    remove certbot</tt>, or <tt>sudo yum remove certbot</tt>.
+    common examples are <code>sudo apt-get remove certbot</code>, <code>sudo dnf
+    remove certbot</code>, or <code>sudo yum remove certbot</code>.
     </p>
     <p>
     If you previously used Certbot through the certbot-auto script, you should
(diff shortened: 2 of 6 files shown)


_scripts/instruction-widget/templates/install/pip.html
```diff
@@ -0,0 +1,81 @@
...
+{{#advanced}}
+{{#dns_plugins}}
+{{/dns_plugins}}
+{{/advanced}}
+
+{{> dnspluginssetup}}
+
+<li>
+  [Optional] Upgrade certbot
+  <p>When you're ready to upgrade certbot, run the following command on the command line on the machine.
```
bmw, line 74: nit: I'd personally suggest a little more strongly people keep Certbot up-to-date. What do you think about something like removing `[Optional]` and saying something like "It's important to occasionally update Certbot to keep it up-to-date. To do this, run the following command on the command line on the machine."?

_scripts/instruction-widget/templates/install/pip.html
```diff
@@ -0,0 +1,81 @@
...
+  Execute the following instructions on the command line on the machine to set up a virtual
+  environment.
+  <pre class="no-before"><ol>
+        <li>sudo python3 -m venv /opt/certbot/</li>
+        <li>{{install_command}} --upgrade pip</li>
+  </ol></pre>
+  </p>
+</li>
+
+{{> installcertbot}}
```
bmw, line 51: Like we do in our snap instructions, somewhere above this point I think we may want to encourage people to uninstall any Certbot OS packages. What do you think?

bmw (commented):
> Questions:

For all of these I went ahead and shared my opinion, however, I'm also curious about what you think as I think you have a good instinct for this kind of stuff.

> Should we give people the command to upgrade certbot monthly via cron and remove [Optional]?

I personally don't think so. I think the chance of things breaking when upgrading is too high. For example, `cryptography` could require a newer version of rust or dropped support for the user's version of OpenSSL.

> Should we switch to recommending people blow away the venv each time?

I wrote more about this on the Google doc, but I personally think we should either just recommend `pip install --upgrade` or blowing away the venv. If we feel the need to document both, then I think `pip install --upgrade` maybe isn't reliable enough and we should just recreate the venv.

> Should the upgrade instruction be down below with renewal, so people can do the most important things first? It's a little more work (it'll need to be moved into a different partial) but super doable.

I think that'd be nice!

> We don't mention third-party plugins elsewhere; should we remove that from here?

I think so. I think on the default tab we shouldn't have this step but on the wildcard tab we'd have the instructions to install the DNS plugins.

> What should the system name in the uri and the dropdown/title be?

Maybe make the system name in the uri `pip` and the dropdown/title "Other Linux (from source)" while changing the other "Other Linux" option to "Other Linux (snapd)"?

ohemorange:
>> Should we give people the command to upgrade certbot monthly via cron and remove [Optional]?
> I personally don't think so. I think the chance of things breaking when upgrading is too high. For example, cryptography could require a newer version of rust or dropped support for the user's version of OpenSSL.

Ack, that's a good point. Since we're only updating some of the packages and can't be sure people have updated their system packages.

>> Should we switch to recommending people blow away the venv each time?
> I wrote more about this on the Google doc, but I personally think we should either just recommend pip install --upgrade or blowing away the venv. If we feel the need to document both, then I think pip install --upgrade maybe isn't reliable enough and we should just recreate the venv.

Hm, looking at your comments on the Google doc. Did certbot-auto actually blow away the venv every time? I thought we had eventually fixed it somehow so it didn't have to. I know we often used to run into issues while testing, but that's probably because we were only updating certbot's source directly by pulling from master, which certainly wouldn't pull in any necessary updates to dependencies.

>> Should the upgrade instruction be down below with renewal, so people can do the most important things first? It's a little more work (it'll need to be moved into a different partial) but super doable.
> I think that'd be nice!

Cool, I will do this.

>> We don't mention third-party plugins elsewhere; should we remove that from here?
> I think so. I think on the default tab we shouldn't have this step but on the wildcard tab we'd have the instructions to install the DNS plugins.

Great, that's where I was leaning as well.

>> What should the system name in the uri and the dropdown/title be?
> Maybe make the system name in the uri pip and the dropdown/title "Other Linux (from source)" while changing the other "Other Linux" option to "Other Linux (snapd)"?

Hm, I think since we've already been using "pip" for Other Linux's uri for some reason, using pip for this might break urls if anyone's bookmarked or posted it somewhere. I like the idea of using other linux () for both of these. I guess since it's python installing from pip is technically from source, but that feels a little weird. Other Linux (pip)? Other Linux (venv)? I think (pip) is pretty analogous to (snapd). Maybe pip-venv for the uri?

_scripts/instruction-widget/templates/install/pip.html
ohemorange: great call!

_scripts/instruction-widget/templates/install/pip.html
ohemorange: I'd like to indicate that this step shouldn't be done right now. What do you think about [Monthly]? Otherwise, totally agree, great suggestion.

_scripts/instruction-widget/templates/install/pip.html
```diff
@@ -0,0 +1,81 @@
...
+
+    <p>Commands to install system dependencies may look like the following, run on the command
+    line on the machine:
+      <pre class="no-before"><ol><li># For APT-based distributions (e.g. Debian, Ubuntu ...)</li>
+        <li>sudo apt update</li>
+        <li>sudo apt install python3-dev python3-venv libaugeas0</li>
+        <li># For RPM-based distributions (e.g. Fedora, CentOS ...)</li>
+        <li># NB1: old distributions will use yum instead of dnf</li>
+        <li># NB2: RHEL-based distributions use python3X-devel instead of python3-devel (e.g. python36-devel)</li>
+        <li>sudo dnf install python3-devel augeas-libs redhat-rpm-config</li></ol></pre>
```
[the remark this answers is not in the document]
ohemorange: Looks like we needed it for Fedora 23: https://github.com/certbot/certbot/pull/1527

I'll check if it's still needed on Fedora...32......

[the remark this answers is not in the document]
ohemorange: Looks like some time in the past 5 years, this instruction became obsolete; removing it. We should remove this and also the outdated cryptography packages from the contributing docs as well.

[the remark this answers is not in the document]
ohemorange: Made an issue for that https://github.com/certbot/certbot/issues/8696

kallisti5:
excited for this one... the snapd requirement is a bit much for a set of python scripts.  Immediately ran into issues messing around with snapd on systems with low memory that have never had issues before... certbot is literally just a set of python scripts.

_scripts/instruction-widget/templates/install/pip.html
```diff
@@ -23,21 +23,43 @@ <h3>Partial support</h3>
...
-        <li># For RPM-based distributions (e.g. Fedora, CentOS ...)</li>
-        <li># NB1: old distributions will use yum instead of dnf</li>
-        <li># NB2: RHEL-based distributions use python3X-devel instead of python3-devel (e.g. python36-devel)</li>
-        <li>sudo dnf install python3-devel augeas-libs redhat-rpm-config</li></ol></pre>
+        <li>sudo apt install python3 python3-venv libaugeas0</li>
+      </ol></pre>
+    </p>
+    <p>For RPM-based distributions (e.g. Fedora, CentOS ...):
+    <br>NB1: old distributions will use yum instead of dnf
+    <br>NB2: RHEL-based distributions use <code>python3X</code> instead of <code>python3</code>
```
bmw, line 35: nit: I personally think this "NB#: ..." text from https://certbot.eff.org/docs/contributing.html#running-a-local-copy-of-the-client looks kind of sloppy here. If you agree, can we reformat this maybe as a note with more or less complete sentences? If you disagree, feel free to leave it as is.

bmw (commented):
> Hm, looking at your comments on the Google doc. Did certbot-auto actually blow away the venv every time? I thought we had eventually fixed it somehow so it didn't have to. I know we often used to run into issues while testing, but that's probably because we were only updating certbot's source directly by pulling from master, which certainly wouldn't pull in any necessary updates to dependencies.
>
> EDIT 3: wow, we really did just blow away the old venv every time. I still have a little hesitation from a UX perspective, since it'll be a little longer, and now the user is probably actively waiting for it to finish instead of having it run from cron. But I'd much rather mimic existing workflows that we know haven't been actively problematic, rather than having to deal with new unknown issues. So let's blow it away! Whoo!
>
> Edit 4: Ah...just noticed a few things. Easiest is that that rm command needs sudo. Next is that the current instruction doesn't mention the dns packages at all, so if we left it we'd have to fix that. But that leads into finally: the existing installation steps are scattered among various steps: "Set up a Python virtual environment", "Install Certbot", and possibly "Install correct DNS plugin". (Which, further thing, here and possibly also in snapd, "Prepare the Certbot command" should probably be after "Install correct DNS plugin".) What are the chances that people just shoved the DNS credentials into the venv, which we've now told them to delete? We could leave the wording as is, saying "repeat all installation instructions." If we keep the --upgrade, it's not so bad to conditionally put in a certbot-dns-x entry if we're on dns instructions. We could also tell people explicitly which steps to repeat; I don't think we should try to take all those contents and shove them into this step. Leaning towards this last one, though it's a little messy as instructions go. So weighing that against the potential pitfalls of just telling people to run pip --upgrade certbot[s]. Ooh we could even extract the things with certbot in the name from pip freeze and pass those back in, instead of having to conditionally include certbot-dns-x, making the command copy-pastable.

Deleting the venv sure does complicate the instructions doesn't it? I've personally changed my mind and think we should tell people to try `pip install --upgrade` and ideally include `certbot-dns-<PLUGIN>` in the upgrade command on the wildcard tab. I'm somewhat tempted to try just this without the `rm -rf` fallback and see how it goes, but I don't have a strong opinion here. If you want to keep it, I think we should update the command to use `sudo`.

> Hm, I think since we've already been using "pip" for Other Linux's uri for some reason, using pip for this might break urls if anyone's bookmarked or posted it somewhere. I like the idea of using other linux () for both of these. I guess since it's python installing from pip is technically from source, but that feels a little weird. Other Linux (pip)? Other Linux (venv)? I think (pip) is pretty analogous to (snapd). Maybe pip-venv for the uri?
>
> EDIT: actually it turns out "Other Linux" is just our current fallthrough; it's just going to whatever we want the default to be. In light of that, I think I'd like to have a specific pip option much in the same we have a specific snapd option, so people can scroll quickly through to find the pip instructions if they're specifically looking for the pip instructions. And then we can still have two Other Linuxes if we want, do we even need a fallthrough? Why do we even have a fallthrough, it's a dropdown, let's just specifically say where those two other linuxes go.
>
> EDIT 2: Or were you thinking get rid of the plain snapd option and just have Other Linux (pip) and Other Linux (snapd)? I wouldn't want to erroneously indicate that those instructions won't work on systems that are listed in the menu

I'm fine with `pip-venv` for the URI and "Other Linux (pip)" and/or "pip" dropdowns leading to that URI. I'm personally not too worried about the latter because I would like to do a dramatic overhaul of the OS dropdowns in the near future that makes that list much shorter and reduces/eliminates all the redundancy we currently have. Because of this, I think/hope whatever text we choose now is a short term thing.

_scripts/instruction-widget/templates/getting-started/renewal.html
```diff
@@ -73,3 +73,17 @@
...
+<li>
+  [Monthly] Upgrade certbot
+  <p>It's important to occasionally update Certbot to keep it up-to-date. To do this, run the
+    following command on the command line on the machine.
+    <pre class="no-before"><ol>
+        <li>{{install_command}} --upgrade {{package}}</li></ol></pre>
+    </p>
+
+  <p>If this step leads to errors, run <code>rm -rf /opt/certbot</code> and repeat all installation instructions.</p>
+  <p>For automatic upgrades, the command in this step can be run by a monthly cron job.</p>
```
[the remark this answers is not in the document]
ohemorange: Agreed.

_scripts/instruction-widget/templates/install/pip.html
ohemorange: I agree, will give it a try.

ohemorange:
I decided to just go with what the dropdown options probably actually should be, despite all that prevaricating. Lmk if you want me to change it to something else.

Also, here's some updated relevant screenshots:

DNS Upgrade
Screenshot from 2021-03-02 16-58-19

Non-DNS Upgrade
Screenshot from 2021-03-02 16-57-06

RHEL install
Screenshot from 2021-03-02 16-51-59

bmw (approved):
LGTM!

Thanks again for picking this up.

PR #702, opened 2021-03-05
Fix pip typo

bmw:
The parenthetical currently says:

> this might be called python3, python3, or python3-venv

I started to just fix this by talking about `python3X` such as `python36`, but we include that same information in the sentences below. Rather than duplicating the information here, I figured we could just simplify this sentence and keep the system specific details in one place.


alexzorin (approved):
Agreed!

PR #703, opened 2021-03-24
remove duplicate DreamHost entry

jtracey:


alexzorin (approved):
Nicely spotted, thank you for the fix.

PR #707, opened 2021-04-24
link snapd's "alternate instructions" to pip

### LINKED ISSUE #705: Update snapcraft "you can find alternate installation instructions here" link
It should link to our pip instructions directly, since that's our preferred alternate, rather than the information-overload page in the docs.

alexzorin:
Fixes #705.


ohemorange:
I don't actually know the right way to fix this, but I did notice we already pass `--url-ignore '/twitter.com/'`, so maybe  `--url-ignore '/twitter.com/,/lets-encrypt\/pip-/'` will work (those are apparently regexes in there).

alexzorin:
I don't think I'm going to get to this.

Issue #711, opened 2021-05-12
Docs: Search Stuck Indefinitely


zoracon:
Reported by: @jsha 

---

## Steps to reproduce:

    Visit https://certbot.eff.org/docs/search.html?q=acmev2&check_keywords=yes&area=default

The page shows "Searching..." indefinitely, and a JS error in the console:

Uncaught ReferenceError: Stemmer is not defined

    query https://certbot.eff.org/docs/_static/searchtools.js:158

    setIndex https://certbot.eff.org/docs/_static/searchtools.js:98

alexzorin:
Had a go at this in https://github.com/EFForg/sphinx_rtd_theme/pull/6.

bmw:
This is fixed.

Issue #682, opened 2020-12-10
Stop using submodules


bmw:
git submodules are awful. On top of that, we regularly forget to update the submodule for the Certbot website. Is there something more usable we can do instead?

I'm not sure if there's code out there we can use for this on a site like readthedocs, but can we essentially just always build the docs for the latest tag on the Certbot repo with an easy way for us to override that if we want to update something between releases? (Bonus points if the system also helps us to remember to stop overriding the commit used in some way, however, I certainly do not think this is essential.)

bmw:
What I personally think we should do here is to just start hosting our documentation on readthedocs.org. It allows us to stop using submodules, easily toggle through their UI which commit is used for the live documentation, and do things like by default always show the documentation for our latest git tag.

adferrand:
I remember that something like two years ago, there were some PRs to remove last traces of a something looking like a Certbot documentation hosted on readthedoc.

Was it really the case? What was the reason if so to move away from it if so?

bmw:
I think you're remembering https://github.com/certbot/certbot/issues/8155 which was about the letsencrypt readthedocs page that we struggled to make redirect to https://certbot.eff.org/docs.

PR #712, opened 2021-05-19
Remove custom sphinx theme

bmw:
Fixes https://github.com/certbot/website/issues/711.

Removing this theme was originally suggested in https://github.com/EFForg/sphinx_rtd_theme/pull/6. Looking at https://github.com/EFForg/sphinx_rtd_theme/compare/eeff564...master, it seems the main changes made in our fork were to add analytics, a navigation bar at the top, and tweak things like the font/colors to better match the added navigation bar and the rest of the Certbot site. I don't think it's worth continuing to maintain this fork for these reasons though. I think the default readthedocs theme is sufficient and its much easier for us to maintain.

With the non-forked theme, the docs look like:
Screen Shot 2021-05-19 at 4 00 20 PM

### DIFF
--- a/_docs.sh
+++ b/_docs.sh
@@ -12,7 +12,6 @@ case "$1" in
   "install" )
     cd _docs
     source ./venv/bin/activate
-    pip install --upgrade git+https://github.com/EFForg/sphinx_rtd_theme.git
     cd certbot
     make -C docs clean html epub latex latexpdf > /dev/null
     ;;


bmw:
I also successfully tested that this change fixes the search problems.

ohemorange:
This is actually a pretty hard call, imo. Especially the navigation bar and theme -- without it, it looks like you've been taken to a different site, rather than a cohesive part of our website as a whole. That's not the worst thing and it's for sure a tradeoff, but I didn't realize that literally every part of the theming came from this. Is there really no better-supported way to get those features without forking?

bmw:
The only other approach I see would be to try adding separate CSS files that override the theme as described [here](https://sphinx-rtd-theme.readthedocs.io/en/latest/configuring.html#adding-custom-css-or-javascript). Maybe that'd work better, but I worry that this would be quite fragile due to the extent we're customizing the theme here.

Another idea I had to stop using both this custom theme and git submodules is to start hosting our static docs on readthedocs which I suggested [here](https://github.com/certbot/website/issues/682#issuecomment-844554234).

ohemorange (approved):
We are going to merge this and then just modify the css to do similar things after.

PR #714, opened 2021-05-20
Update to Node.js 10

adferrand:
Using the official way proposed to install Node.js on debian/ubuntu systems (https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions), this PR upgrades the build Docker to Node.js 10.

I checked that this version is working with the current set of dependencies (note that Node.js 12 is not).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,15 +36,14 @@ RUN echo 'gem: --no-document' >> /usr/local/etc/gemrc && \
     gem install jekyll html-proofer
 
 # Install node and dependencies
-RUN apt-get install -y npm
-RUN npm install -g n
-RUN n 8.12.0
-RUN npm install -g npm gulp-cli
+RUN curl -fsSL https://deb.nodesource.com/setup_10.x | bash - && \
+    apt-get install -y nodejs && \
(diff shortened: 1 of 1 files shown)


bmw:
I tested this and it works just fine, however, I don't like that we're deviating from the [version of node used in Travis](https://github.com/certbot/website/blob/e93982670c9d987754859f67883d521ec64651c7/.travis.yml#L4) or [documented in our README](https://github.com/certbot/website/tree/e93982670c9d987754859f67883d521ec64651c7#install). We could upgrade those too, but I worry a bit about stepping on our engineering and design teams toes here, especially when they're (at least planning to soon start) working on the site and Node.js 10 has also reached its EOL.

I want to know what you think, but my preference would be to either just stick with node 8 for now or upgrade to node 10 (or later) everywhere and check in with Alexis and/or our E&D team before merging this change. If you want to go the 2nd route, I can help with communication here.

adferrand:
Yes my modification makes less sense if other parts of the workflow relies on Node 8. If all of this is about to be refactored, and given that Node 10 is also EOL anyway, if prefer to cancel this PR.

PR #716, opened 2021-05-27
Improve renewal instructions in the instruction generator

ohemorange:
These changes are done to match our updated [cli ux](https://github.com/certbot/certbot/pull/8860), our updated [docs instructions](https://github.com/certbot/certbot/pull/8870), and feedback from the CISPA Usable Security group. 

pip apache (cron not included, no certonly)
pip-apache

pip other (cron not included, certonly)
pip-other

snapd apache (cron included, no certonly)
ubuntu-no-certonly

snapd other (cron included, certonly)
snapd-certonly

This PR does not include the Windows instructions, which are hard-coded on their own in `windows.html`.

PR #718, opened 2021-05-28
Give both cron-included and non-cron-included users instructions on how to test using --dry-run

ohemorange:
As pointed out [here](https://github.com/certbot/website/pull/716#pullrequestreview-670978337), we weren't showing users who needed to set up automatic renewal how to test their renewal setup. We should do that. Also, now that we always have users use directory hooks, we should have them test after they've set up the hooks.

Snap, apache (no certonly)
snap-apache2

Snap, none of the above (certonly)
snap-other2

Snap, haproxy (certonly)
snap-haproxy

Pip, apache (no certonly)
Screenshot from 2021-05-28 14-26-59

Pip, none of the above (certonly)
none-of-the-above

Pip, haproxy (certonly)
Screenshot from 2021-06-01 16-46-23

### DIFF
--- a/_scripts/instruction-widget/get-started.js
+++ b/_scripts/instruction-widget/get-started.js
@@ -44,6 +44,7 @@ module.exports = function(context) {
         context.imperative = "you should probably"
         template = "haproxy";
         context.certonly = true;
+        context.haproxy = true;
     }
 
     plesk_getting_started = function() {
--- a/_scripts/instruction-widget/get-started.js
+++ b/_scripts/instruction-widget/get-started.js
@@ -43,6 +43,7 @@ module.exports = function(context) {
         context.officially = "officially ";
         context.imperative = "you should probably"
         template = "haproxy";
+        context.certonly = true;
     }
 
     plesk_getting_started = function() {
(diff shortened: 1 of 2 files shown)


alexzorin (commented):
1. Given that our example use case for directory hooks is haproxy, it's a little odd that selecting `haproxy/pip` or `haproxy/snapd` doesn't show these instructions. Probably just due to the design of the generator.
2. Are we putting the `--dry-run`  instructions below the "directory hooks" instructions intentionally? My first reaction was  that it takes a lot of reading to get to it. On the other hand, if it's important for the user to read all that, maybe it's fine as it is.

ohemorange:
1. Oh no, what?? Yeah that looks like a bug. And we probably should fix Windows as well as some point, now that I'm looking through all our special-cased instructions that aren't just pulling from the file I updated here...

2. Yes, that was intentional. My thought was that if they're testing, they should have it as close as possible to the final setup, including the hooks. I wasn't thinking about importance though, that's a good point. If we swap it, we probably want to tell them to run `--dry-run` again, what do you think?

ohemorange:
Well, now it looks a little weird that we're saying "if your webserver is haproxy" when we do know it is, I can fix that.

ohemorange:
travis??

ohemorange:
Updated screenshots in the top picture.

alexzorin (approved):
I think this is pretty good. 

Good catch with Windows. I created #720 to track that.

Issue #721, opened 2021-06-11
opensuse instructions need replacing


zoopla-jb:
The opensuse instructions use snap which is completely unnecessary and is not supported in opensuse. This is true for Leap and Tumbleweed
 
This one step would replace steps 2-6 on this [page](https://certbot.eff.org/lets-encrypt/tumbleweed-nginx) :

2. install certbot with `sudo zypper in python3-certbot`

Any additional plugins can be found by typing `zypper se certbot`.

I would also suggest renaming Leap 15.0 to Leap 15.x, as 15.0 and 15.1 are end of life, whilst 15.2 and 15.3 are still active.

bmw:
Using snap may not be supported by the opensuse maintainers, however, there are instructions from the snap team for installing snap on opensuse at https://snapcraft.io/docs/installing-snap-on-opensuse.

We have chosen to recommend snaps wherever possible because it ensures people are able to a Certbot package that stays up-to-date and does things like configures automated renewal by default. If you don't want to use snaps, there are other options available, but we're planning to continue to recommend snaps in our instructions here so I'm going to close this issue.

Thanks for pointing out the Leap version number. I've opened https://github.com/certbot/website/pull/722 to fix that.

PR #722, opened 2021-06-11
Change openSUSE Leap dropdown version to 15.x

bmw:
15.0 has reached its EOL, but there are still other openSUSE Leap 15 versions so let's use a more generic version number here. More info about openSUSE's versions can be found at https://en.wikipedia.org/wiki/OpenSUSE_version_history#Version_history and this suggestion came from https://github.com/certbot/website/issues/721.

A screenshot with this change is:
Screen Shot 2021-06-11 at 9 42 57 AM

### DIFF
--- a/_data/inputs.json
+++ b/_data/inputs.json
@@ -165,10 +165,10 @@
           "version": "0"
         },
         {
-          "name": "openSUSE Leap 15.0",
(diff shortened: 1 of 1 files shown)


ohemorange (approved):
lgtm

PR #729, opened 2021-08-30
macos: use $(command -v certbot) for cron

alexzorin:
Certbot is not in the PATH when installed from Homebrew.

---

Per https://github.com/certbot/certbot/issues/8974.

Excuse the templating. I couldn't work out a way to do this in the "controller" without refactoring a bunch of other stuff. Any bright ideas on not having this be so hacky would be appreciated.

After this change, the macOS command becomes:

    echo "0 0,12 * * * root python -c 'import random; import time; time.sleep(random.random() * 3600)' && $(command -v certbot) renew -q" | sudo tee -a /etc/crontab > /dev/null

and the Linux (e.g. `pip`) command remains as:

    echo "0 0,12 * * * root /opt/certbot/bin/python -c 'import random; import time; time.sleep(random.random() * 3600)' && certbot renew -q" | sudo tee -a /etc/crontab > /dev/null

### DIFF
--- a/_scripts/instruction-widget/install.js
+++ b/_scripts/instruction-widget/install.js
@@ -234,6 +234,7 @@ module.exports = function(context) {
     template = "macos";
     context.base_command = "certbot";
     context.install_command = "brew install";
+    context.cron_base_command = "$(command -v certbot)";
   }
 
   // This function is currently unused, but we keep it around to make it easy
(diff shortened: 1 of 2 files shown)


bmw (approved):
LGTM!

I personally think the way this was written is pretty reasonable.

The best alternative that immediately comes to mind would be to always set `cron_base_command` in our code maybe by adding some logic somewhere that sets it equal to `base_command` if it's not set to something else, but I'm personally not worried about it. If you'd like to tweak the code here or talk about this more, I'm happy to take a look.

Issue #487, opened 2019-10-19
Stop using Vivian's GitHub account to publish website changes


bmw:
It works great, but since she's no longer at EFF we should probably change it.

I propose using the @certbotbot whose credentials are shared amongst the Certbot team and is currently being used to automate builds on Docker Hub.

PR #730, opened 2021-08-31
Switch to GitHub Actions

bmw:
Fixes https://github.com/certbot/website/issues/487 and Travis not running because [it's no longer free for open source projects](https://travis-ci.community/t/org-com-migration-unexpectedly-comes-with-a-plan-change-for-oss-what-exactly-is-the-new-deal/10567).

This GH Action setup should essentially be the same as our Travis setup except the SSH key being used is one I created for the GH account https://github.com/certbotbot and notifications about failed builds are sent directly to Mattermost.


alexzorin (approved):
LGTM

PR #734, opened 2021-09-01
Fix Mattermost notifications

bmw:
This PR fixes a couple things:

1. `GITHUB_WORKFLOW_URL` is not a valid environment variable. The new approach creates a valid URL which you can see [here](https://github.com/certbot/website/runs/3488055196?check_suite_focus=true#step:3:9).
2. The previous code would also [report errors on pull requests not from forks](https://github.com/certbot/website/runs/3487706069?check_suite_focus=true) and pushes to non-master branches. You can see notifications being skipped on a non-master branch [here](https://github.com/certbot/website/actions/runs/1191396526) and being posted when the conditional is modified so the branch name matches [here](https://github.com/certbot/website/actions/runs/1191405565).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,7 @@ jobs:
         run: |
           WORKFLOW_RUN_URL="$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID"
           echo "{\"text\":\"** :warning: $GITHUB_REPOSITORY: Build failed :warning: ** | [(see details)]($WORKFLOW_RUN_URL) \"}" > mattermost.json
+          cat mattermost.json
         if: ${{ failure() }}
           #- uses: mattermost/action-mattermost-notify@master
           #  env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,7 @@ jobs:
           submodules: true
       - name: Install OS packages
         run: |
+          geqhgjkeqgheq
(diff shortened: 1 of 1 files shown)


bmw:
> There seems to be some kind of portability issue with `htmlproofer` blowing up and locales, if UTF-8 is not available or chosen. It only comes up when trying to run the workflow locally with e.g. `act`, though.

Thanks for mentioning this. It doesn't seem to be a problem in GH Actions (now?) and this is [handled in our Dockerfile](https://github.com/certbot/website/blob/9b003cce63c96c532dd4baf68b95d09953b0cb8d/Dockerfile#L10-L13), but if you think we should do something else to deal with this, by all means propose/write a fix.

> [The way `gulp` is being called](https://github.com/certbot/website/blob/1add3cb5012e88ffe94b3821748d3a3562b4e762/.github/workflows/ci.yml#L39) seems to be running the `gulp` that comes pre-installed on the GitHub Actions image. To avoid future surprises, we could be doing `npm install -g gulp`, or `npx gulp build` to use the one we're installing, or simply using the preinstalled global one and not installing `gulp` at all. 

I played around with this. My understanding is we're using the preinstalled version of `gulp-cli` but using our own install of `gulp`. This is based on output like [this](https://github.com/certbot/website/runs/3499155707?check_suite_focus=true#step:10:1). Does that match your understanding? If so, maybe we should do a (global?) `gulp-cli` install and then continue to install `gulp` locally (maybe by just running `npm install` assuming that works).

I'm merging this for now though because like you said this is unrelated to the fix here.

PR #737, opened 2021-09-22
windows: link to alternative clients is outdated

webprofusion-chrisc:
The current link to alternative clients is to a community discussion that's not maintained. This change updates the link to point to a maintained document.

The alternative would be to remove the link.


bmw (approved):
Thanks!

PR #741, opened 2021-09-30
ci: only run on 'push' for the master branch

alexzorin:
In https://github.com/certbot/website/pull/739, it looks like the CI tried to push a website build (which thankfully failed due to a missing secret).

It turns out the `push` event is triggered for any branch, not just the default branch. When @dependabot pushed to the https://github.com/certbot/website/tree/dependabot/npm_and_yarn%2Fdatatables.net-1.11.3 branch, the "Push build" part of the workflow was triggered.

This change limits the `push` event to the `master` branch only. It shouldn't affect PRs because supposedly the `pull_request:synchronize` event will be re-triggered every time there are new commits in the PR's source branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,8 @@
 name: Certbot website CI
 on:
   push:
+    branches:
+      - master
   pull_request:
   schedule:
     # Run at 4pm UTC or 9am PST


bmw:
What do you think about https://github.com/certbot/website/pull/742? My main concern about this approach is it breaks https://github.com/certbot/website#building-with-github-actions.

alexzorin:
Thanks, I'd totally missed that functionality.

PR #759, opened 2022-01-18
Remove all unused files, now that the website has been migrated

ohemorange:
Fixes #756.

CI requirements have been turned off for this repo.

What's left after this:

```
~/website$ tree -a -L 2
.
├── _data
│   └── hosting_providers.json
├── .git
│   ├── branches
│   ├── COMMIT_EDITMSG
│   ├── config
│   ├── description
│   ├── FETCH_HEAD
│   ├── HEAD
│   ├── hooks
│   ├── index
│   ├── info
│   ├── logs
│   ├── modules
│   ├── objects
│   ├── ORIG_HEAD
│   ├── packed-refs
│   └── refs
├── .gitignore
└── README.md

9 directories, 11 files
```


alexzorin (changes requested):
What do you think about doing something with the README? 

- We could replace most of the content with short note stating that the website is no longer managed through the repository.
- We could have advice for users who want to suggest content changes, maybe pointing them to Mattermost.
- The *Hosting providers list* section is still relevant and should stay.

ohemorange:
You're correct, we should do something about the readme. This PR came about during the group triage as a direct response to the dependabot issue, with the idea of looking at the readme later. The changes you suggested don't actually seem to be that bad, but I think I'd rather just clean this up and do that separately.

alexzorin (approved):

PR #768, opened 2022-06-02
change do.de to full support

henrikalves:


alexzorin (approved):
That's great news, thank you, LGTM.

henrikalves:
seems like the data on the site was never updated

PR #777, opened 2022-10-17
Add Koumbit to hosting_providers.json

johnpink:


alexzorin (approved):
Thanks for fixing up the JSON as well.

PR #792, opened 2023-06-11
Include avantiv.com in hosting providers list

rmartinsjr:


alexzorin:
To clarify the "full" support: when users sign up to these hosting plans, do they need to do anything to enable SSL certificates? Like going to the user interface and pressing a button? Or is it done automatically in the background?

rmartinsjr:
Nothing is required to enable SSL. It's already provisioned with SSL support.

alexzorin (approved):
Thanks!

PR #813, opened 2024-08-31
Update hosting_providers.json

nikhildevre:


wgreenberg (changes requested):
it looks like Namesilo does provide SSL certificates, so why is this categorized as "no"?

wgreenberg (approved):
i can merge from this branch, thanks!

PR #819, opened 2024-11-05
Added regfish to hosting_providers.json

amallek:


wgreenberg:
Sorry if I'm wrong (I don't speak German and had to use Google Translate), but it seemed like https://regfish.de/zertifikate suggests that certificates aren't free?

PR #820, opened 2024-11-23
Create southerndatacenter

Pronoor:
Hosting service Provider with Free SSL certificates for dedicated and shared hosting use.


Pronoor:
Thank you

wgreenberg (changes requested):
this isn't how we add providers to the site. please look at https://github.com/certbot/website/blob/master/_data/hosting_providers.json and create an entry in that JSON file w/ the required fields

PR #831, opened 2025-03-17
Add Framer

ankon:
The `tutorial` points to the article spelling out that Framer will always configure HTTPS, automatically.

I added the entry to the end, because the file seems to be unsorted. Let me know if it should be moved to a different place.

### DIFF
--- a/_data/hosting_providers.json
+++ b/_data/hosting_providers.json
@@ -2808,5 +2808,15 @@
     "plan": "",
     "reviewed": "2024.8.16",
     "note": "This website is in Dutch."
+  },
+  {
+    "name": "Framer",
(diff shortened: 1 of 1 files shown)


ohemorange (changes requested):
You're correct, it's unsorted, so the placement is good. It mostly looks good, except that "tutorial" is for steps a user would have to take to turn on https specifically, whereas "announcement" is for confirmation that it's automatic and the user doesn't have to do anything extra. I think therefore that "announcement" would be the most correct place for the link.

ankon:
Sounds good, updated in https://github.com/certbot/website/pull/831/commits/4525cb1567da3389f60ef1c136f34fc28e379309

PR #840, opened 2025-06-27
[TOOLING] Create host.yaml

zoracon:
Figured it might be useful for future requests. Let me know if it's not really helpful.

### DIFF
--- a/.github/ISSUE_TEMPLATE/host.yaml
+++ b/.github/ISSUE_TEMPLATE/host.yaml
@@ -1,0 +1,57 @@
+name: Add a Host
+description: Adds a new host to our Hosting Providers list
+title: "[New Host]: "
+body:
+  - type: input
+    id: name
+    attributes:
+      label: Name
+      description: Name of host provider
+    validations:
+      required: true
+  - type: input
+    id: link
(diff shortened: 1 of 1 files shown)


ohemorange:
I mean, ideally people will open PRs instead of issues, no?

zoracon:
> I mean, ideally people will open PRs instead of issues, no?

@ohemorange I thought about just a PR template with the structure someone would need as an example snippet. What do you think?

zoracon:
> I guess that could work? What problem is that solving? I feel like the json structure makes the basic format clear, and the [README](https://github.com/certbot/website/blob/master/README.md), which literally shows up on the main page of the repo, explains what the fields are. Like I guess we could put "have you considered reading the README and opening a PR based on that" but I don't think we encourage a process where we have to manually create a PR from inputs.

It's not necessarily a rampant problem. Just sometimes I see missing fields in PRs pop up. And figured a template would smooth this over and make it even more clear.

ohemorange:
Hm, maybe a PR template with a checklist based on the info from the readme then?

ohemorange:
Closing in favor of #854

PR #843, opened 2025-08-05
Clean the hosting providers file

ohemorange:
Make sure that recent additions are [to spec](https://github.com/certbot/website/?tab=readme-ov-file). Scan through notes to remove egregious advertising in the notes field that doesn't provide additional information.

### DIFF
--- a/_data/hosting_providers.json
+++ b/_data/hosting_providers.json
@@ -6,7 +6,7 @@
     "tutorial": "",
     "announcement": "https://www.theory7.net/webhosting/",
     "plan": "",
-    "reviewed": "2025.02.03",
+    "reviewed": "2025.08.05",
     "note": "Free & automated SSL for all webhosting and domains"
     },
    {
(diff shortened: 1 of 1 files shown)


ohemorange:
I agree that can be made clearer and will clear that up in a separate PR.

ohemorange:
wow, we really stopped being particular about plan vs announcement at some point. still worth fixing it going forward I think -- initially, plan was for a link with plans to implement more support that weren't there yet. might honestly be better to get rid of it at this point and just do something like tutorial and full support evidence, though I wouldn't want to make more work for e&d by changing it.

bmw:
i personally kinda like the idea of consolidating announcement and plan. what if we changed all/most current plan links to announcements and made the following changes to the instructions:

* say full should have an announcement and "shouldn't need a tutorial to turn on https. an exception might include instructions of what to do if something goes wrong and the automatic https doesn't work."
* say plan is no longer used and should be left as ""

this allows us to consolidate the two fields in a way that definitely doesn't require anything from e&d

also, imo bonus points should be given if we move any text we want from https://github.com/certbot/certbot/wiki/Triage-instructions#reviewing-hosting-providers-requests to the readme and we just link to the readme in that wiki entry. deduplicating the two seems nice to me

whatcha think?

ohemorange:
I like all of that. tbh I always forget the wiki link is there, and just use the readme.

ohemorange:
issues made

